PR #31, opened 2024-03-05
IR-147 Add Incident Reporting domain model and related classes 

Mjwillis:
Added new domain model classes representing various entities for reporting incidents in a prison setting. These include main classes like IncidentReport, and supporting classes like IncidentLocation, IncidentCorrectionRequest, and Evidence. Also, corresponding database migration script was added. This sets up the foundation for the incident reporting feature in the HMPPS service.


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/HistoricalIncidentResponse.kt
```diff
@@ -0,0 +1,49 @@
...
+@Entity
+class HistoricalIncidentResponse(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long? = null,
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  val incident: IncidentReport,
+
+  val dataPointKey: String,
```
ushkarev, line 23: minor point on naming… is “data point” a little too generic? as in, can data points be grouped together into clusters of some sort? tbh, at this point we don’t know what we’re making so i have no better terminology

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentLocation.kt
```diff
@@ -0,0 +1,24 @@
...
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long? = null,
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  val incident: IncidentReport,
+
+  val locationType: String,
+
+  val locationId: String,
```
ushkarev, line 21: is this a nomis and/or prison-api id?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentCorrectionRequest.kt
```diff
@@ -0,0 +1,27 @@
...
+import jakarta.persistence.Entity
+import jakarta.persistence.FetchType
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.ManyToOne
+import java.time.LocalDateTime
+
+@Entity
+class IncidentCorrectionRequest(
```
ushkarev, line 12: do you think this will need to have a historical vs. modern (/new/whatever we call the future) version?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentType.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.incidentreporting.jpa
+
+enum class IncidentType {
```
ushkarev, line 3: not asking for this to be changed, but i figure this too would need to be an entity in future…?

ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/HistoricalIncidentResponse.kt
Mjwillis: Yes happy to change - just don't know to what!

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentType.kt
Mjwillis: Maybe?  But introducing a new incident type would require a lot of coordination from different teams, not sure it helps being a table as enum allow us to apply business rules or understand what the types are, anyway hard to know at this point!

ushkarev: i was under the impression that the super-admins can modify report types in nomis… if we can get away with _not_ doing that YAY

PR #34, opened 2024-03-07
IR-169 Remove custom OAuth2 configurations, add SubjectAccessRequest implementation (DUMMY)

Mjwillis:
Deleted custom OAuth2 configurations such as AuthAwareTokenConverter.kt, ResourceServerConfiguration.kt, and WebClientConfiguration.kt, as they are superseded by the HMPPS Kotlin Spring Boot Starter lib. Renamed IncidentReportRepository.kt's path to be nested under the repository package. Added a SubjectAccessRequestService and a corresponding repository for incident reports, allowing authorized entities to fetch data between specific dates.


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/SubjectAccessRequestService.kt
```diff
@@ -0,0 +1,35 @@
...
+@Service
+class SubjectAccessRequestService(
+  private val prisonerInvolvementRepository: PrisonerInvolvementRepository,
+) : HmppsPrisonSubjectAccessRequestService {
+  override fun getPrisonContentFor(prn: String, fromDate: LocalDate?, toDate: LocalDate?): HmppsSubjectAccessRequestContent {
+    val incidentReports = prisonerInvolvementRepository.findAllByPrisonerNumber(prn)
+    val content = incidentReports
+      .filter { it.incident.reportedDate.isAfter(fromDate?.atStartOfDay()) && it.incident.reportedDate.isBefore(toDate?.atStartOfDay()) }
+      .map {
+        IncidentDTO(
```
Mjwillis, line 19: This is just a placeholder

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/SubjectAccessRequestService.kt
```diff
@@ -0,0 +1,35 @@
...
+          reportedDate = it.incident.reportedDate,
+        )
+      }
+    return HmppsSubjectAccessRequestContent(
+      content = content,
+    )
+  }
+}
+
+data class IncidentDTO(
```
ushkarev, line 31: might be worth making it clear (via comment of class name) that this is only specific to information we need to export for SARs

ushkarev (approved):

PR #39, opened 2024-03-13
IR-68: Copy OpenAPI fixes from incentives

ushkarev:
`securitySchemes` in OpenAPI output had a few issues…
1) `in` and `name` only apply to “apiKey” type (not the “http” that we use)
2) `openIdConnectUrl` only applies to “openIdConnect” type (not the “oauth2” that we use)
See https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#fixed-fields-23

[Copied from incentives-api](https://github.com/ministryofjustice/hmpps-incentives-api/pull/549/commits/9f943c4cadbb81a3d176c67a6157d511bfd555b1)


Mjwillis (approved):
LGTM

PR #40, opened 2024-03-14
IR-155 Add functionality for Incident Reports

Mjwillis:
The code introduces the ability to get, create new incident reports.
Support migration of NOMIS incident reports and updates.

*** NOT COMPLETE mapping yet ***

Initial baseline of functionality


src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/MigrateResourceIntTest.kt
```diff
@@ -0,0 +1,210 @@
...
+  lateinit var repository: IncidentReportRepository
+
+  lateinit var existingNomisIncident: IncidentReport
+
+  @BeforeEach
+  fun setUp() {
+    repository.deleteAll()
+
+    existingNomisIncident = repository.save(
+      buildIncidentReport(incidentNumber = "$INCIDENT_NUMBER", reportTime = LocalDateTime.now(clock), source = InformationSource.NOMIS),
```
ushkarev, line 47: should the incident number here be a string with `IR-0000000112414323` to be more realistic?

ushkarev (commented):
but otherwise all the logic makes sense for now :)

PR #47, opened 2024-03-28
IR-183 Add the concept of an event to the incident report.

Mjwillis:
An incident can be linked to an event and an event can have multiple incidents

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentEvent.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentEvent.kt
@@ -1,0 +1,64 @@
+package uk.gov.justice.digital.hmpps.incidentreporting.jpa
+
+import jakarta.persistence.CascadeType
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.FetchType
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.OneToMany
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import uk.gov.justice.digital.hmpps.incidentreporting.dto.EventDetail
+import java.io.Serializable
+import java.time.LocalDateTime
+
+@Entity
+class IncidentEvent(
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long? = null,
+
+  @Column(nullable = false, unique = true, length = 25)
+  val eventId: String,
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentReport.kt
```diff
@@ -40,13 +41,17 @@ class IncidentReport(
...
+  var summary: String?,
   var incidentDetails: String,
 
   val reportedBy: String,
   val reportedDate: LocalDateTime,
   @Enumerated(EnumType.STRING)
   var status: IncidentStatus = IncidentStatus.DRAFT,
 
+  @ManyToOne(fetch = FetchType.LAZY, cascade = [CascadeType.ALL])
+  val event: IncidentEvent? = null,
```
ushkarev, line 53: what do you think about making this non-nullable? it'd make the data model simpler i think. we could just have the typical case being 1 event ↔︎ 1 incident

ushkarev: the statsy reports / consumers of our data wouldn't have to have branching for orphaned incidents vs those grouped into events

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentEvent.kt
```diff
@@ -0,0 +1,64 @@
...
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long? = null,
+
+  @Column(nullable = false, unique = true, length = 25)
+  val eventId: String,
+  val eventDateAndTime: LocalDateTime,
+  val prisonId: String,
+  var summary: String,
+  var eventDetails: String,
```
ushkarev, line 29: what's the difference between summary and details?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentReport.kt
Mjwillis: Yes could do that - its just I need to create one for all NOMIS migrated records

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/IncidentEvent.kt
Mjwillis: Yes not sure there would be one, I've kept it as NOMIS has the concept of title and description and I've reproduced for Event.  Not sure its worth it

karenmillermoj (approved):
lgtm

PR #62, opened 2024-04-26
IR-207: Refactor JPA data model

ushkarev:
This renames JPA entity classes to:
• drop “incident” from their names – which appeared in some, but not all, and is a redundant concept given everything here is about incidents.
• have “question”-“response” pairs making up ”reports” that are grouped into “events” – this will match the designs and NOMIS model more closely

Some database fields have been renamed to more clearly describe their function. Eg. “dataItem” & “dataItemDescription” that form a question are now “code” and “question”.

No logic has changed.

Flyway migrations have been regenerated from scratch so existing databases will need a manual reset before deployment.


Mjwillis (approved):
LGTM

xoen (commented):
Nice work :)

PR #64, opened 2024-04-26
IR-196: Map JPA entities to DTOs

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/Mapping.kt
```diff
@@ -0,0 +1,220 @@
...
+  val report = Report(
+    incidentNumber = "$incidentId",
+    type = mapIncidentType(type),
+    incidentDateAndTime = incidentDateTime,
+    prisonId = prison.code,
+    title = title ?: "NO DETAILS GIVEN",
+    description = description ?: "NO DETAILS GIVEN",
+    reportedBy = reportingStaff.username,
+    reportedDate = reportedDateTime,
+    status = mapIncidentStatus(status.code),
```
xoen, line 26: As mentioned offline, I think it would be _**really nice**_  if the mapping logic were attached to the enums, e.g.

```Kotlin
type = Status.fromCode(status.code)
// type = Type.fromCode(type)
// etc...
```

ushkarev: so you want all these string to enum conversions in the same files as the enum classes themselves?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/Report.kt
```diff
@@ -8,33 +8,57 @@ import java.util.UUID
...
+  @Schema(description = "When the incident report was created", required = true, example = "2024-04-29T12:34:56.789012")
   val reportedDate: LocalDateTime,
+  @Schema(description = "The current status of this report", required = true)
   val status: Status,
+  @Schema(description = "Optional user who this report is currently assigned to", required = false, defaultValue = "null")
   val assignedTo: String? = null,
 
+  @Schema(description = "The question-response pairs that make up this report", required = true)
   val questions: List<Question> = emptyList(),
+  @Schema(description = "Prior versions of this report, created when the report type changes", required = true)
```
xoen, line 39: I know this is not there yet but would a new version be created when anything changes? not just the type no?

ushkarev: apparently only when the type is changed; to match nomis

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/HistoricalQuestion.kt
```diff
@@ -123,4 +125,10 @@ class HistoricalQuestion(
   override fun toString(): String {
     return "HistoricalQuestion(code='$code', responses=$responses)"
   }
+
+  fun toDto() = HistoricalQuestionDto(
+    code = code,
+    question = question,
+    responses = responses.map { it.toDto() },
```
xoen, line 132: I like the `responses = responses.map(HistoricalResponse::toDto),` form better because it makes it much clearer what are the types you're dealing with.

xoen (approved):
LGTM 👍 

(it would be nice to have a `Entity` suffix added to all JPA entities but I guess it could be done in a separate PR)

PR #67, opened 2024-05-03
Removed unnecessary `Serializable` marker interface from JPA entities

xoen:
We don't manually serialize JPA entities to files or over-the-wire and tests pass so these seems to be unnecessary.

(if in the future we find that these are required we can easily readd)


Mjwillis (approved):
lgtm

ushkarev:
[notes down who to blame when everything breaks… 😉]

PR #74, opened 2024-05-09
IR-199: Test NOMIS→Entity→DPS mappings

ushkarev:
Most mappings are already tested as part of end-to-end / integration resource and service tests (there are many JSON input / response assertions).

These extra test cases are for covering edge cases. And, potentially, for future additions of other edge cases.


Mjwillis (approved):
LGTM

PR #80, opened 2024-05-14
IR-224: NOMIS sync endpoint updates report

xoen:
Report is now updated with information provided in sync request:
- type
- staff involvement
- prisoners involvement
- correction requests (called "requirements" in the NOMIS sync request)
- questions/answers
- history

Logic is very similar to one in Mapping. That's why I've extracted it into extension functions that they're reused in the (JPA) `Report`'s `updateWith()` method.
The main difference is that the `updateWith()` method _**clears**_ the associated records before adding whatever is provided by the sync request (using the `Mapping.kt` extensions function.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
@@ -22,6 +22,7 @@ import uk.gov.justice.digital.hmpps.incidentreporting.constants.StaffRole
 import uk.gov.justice.digital.hmpps.incidentreporting.constants.Status
 import uk.gov.justice.digital.hmpps.incidentreporting.constants.Type
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.nomis.NomisReport
+import uk.gov.justice.digital.hmpps.incidentreporting.dto.nomis.addNomisCorrectionRequests
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.nomis.addNomisPrisonerInvolvements
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.nomis.addNomisStaffInvolvements
 import java.time.Clock
@@ -265,6 +266,9 @@ class Report(
 
     prisonersInvolved.clear()
     addNomisPrisonerInvolvements(upsert.offenderParties)
+
+    correctionRequests.clear()
+    addNomisCorrectionRequests(upsert.requirements)
 
     // TODO: need to compare and update other fields and related entities
   }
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -253,7 +244,22 @@ class Report(
       addStatusHistory(newStatus, now, updatedBy)
     }
 
-    // TODO: need to compare and update other fields and related entities
```
ushkarev, line 256: there are still a few fields on `Report` that ought to be copied across.

the slightly harder question is what do we do about `Event`? if the report originally came from NOMIS, the event is a copy of a bunch of report fields – if we know this update is to a report from NOMIS, we should copy across the changes to the event. but what if the original report came from DPS?

xoen: You answered your own question: I don't think the event should be updated when updating/syncing from NOMIS. The upsert itself doesn't contain any (explicit) information about the event and we only populated it on creation with good enough (guessed) values...but after the Report is created the Event could have changed in the DPS service and we wouldn't want an upsert without explicit event information to reset these potential changes.

xoen: Just to recap, for now I'll update the event fields that we update. In the future we may or may not break this link on update.

https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/80/commits/034ee1a850e60738311749476142112959889415

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/NomisSyncResourceTest.kt
```diff
@@ -756,7 +756,15 @@ class NomisSyncResourceTest : SqsIntegrationTestBase() {
           id = existingNomisReport.id,
           incidentReport = syncRequest.incidentReport.copy(
             incidentId = INCIDENT_NUMBER,
+            title = "Updated title",
```
xoen, line 759: I'm explicitly passing these fields to test that the upsert changes them.

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/NomisDtoToJpaMappingEdgeCaseTest.kt
```diff
@@ -156,8 +156,8 @@ class NomisDtoToJpaMappingEdgeCaseTest {
 
       val eventEntity = existingReportEntity.event
       // TODO: changes not copied to event, should they be? probably not (then keep assertions but drop comment)
```
ushkarev, line 158: can remove TODO

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/NomisDtoToJpaMappingEdgeCaseTest.kt
```diff
@@ -175,8 +175,8 @@ class NomisDtoToJpaMappingEdgeCaseTest {
...
```
ushkarev, line 177: can remove TODO

ushkarev (approved):

PR #81, opened 2024-05-15
IR-234: Add paginated endpoint to return all reports

ushkarev:
Filtering is not currently supported, but this will still be useful for periodic jobs to reconcile reports against NOMIS.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
@@ -4,6 +4,10 @@ import com.microsoft.applicationinsights.TelemetryClient
 import jakarta.validation.ValidationException
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import org.springframework.data.domain.Page
+import org.springframework.data.domain.PageRequest
+import org.springframework.data.domain.Pageable
+import org.springframework.data.domain.Sort
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.incidentreporting.config.AuthenticationFacade
@@ -29,6 +33,13 @@ class ReportService(
 ) {
   companion object {
     val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun getReports(
+    pageable: Pageable = PageRequest.of(0, 20, Sort.by("incidentDateAndTime").descending()),
+  ): Page<ReportDto> {
+    return reportRepository.findAll(pageable)
+      .map { it.toDto() }
   }
 
   fun getReportById(id: UUID): ReportDto? {
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
@@ -126,9 +126,10 @@ class ReportResourceTest : SqsIntegrationTestBase() {
               "size": 20,
               "numberOfElements": 0,
               "totalElements": 0,
-              "totalPages": 0
+              "totalPages": 0,
+              "sort": ["incidentDateAndTime,DESC"]
             }""",
-            false,
+            true,
           )
       }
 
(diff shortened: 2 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
```diff
@@ -31,6 +35,13 @@ class ReportService(
     val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
+  fun getReports(
+    pageable: Pageable = PageRequest.of(0, 20, Sort.by("incidentDateAndTime").descending()),
```
Mjwillis, line 39: Is date the best sort or the Incident report ID? I think we want a consistent list order.  i.e. the reports come out in the same order each time.  Could use the created date instead also - its just that incident Date /Time could have a new report added in the middle

Mjwillis (approved):
Comment about the ordering, but happy otherwise

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
ushkarev: this is just the default and sort orders can be specified by the client. i figured that this endpoint will in future be used to answer “gimme the latest reports in MDI” so the date the incident occurred seemed a reasonable default.

the reconciliation process should probably not use a date as it won't result in a stable order (they’re non-unique columns). actually, especially so with created date – we don’t migrate that info from NOMIS so it’s the moment of migration that gets stored. i think that reconciliation should sort on incident number (which works according to the test case) or id.

xoen: I think for the purpose of reconciliation return them sorted by the ID may be the best thing to do. It should be fast as it's the PK and it avoid the scenario where new reports may be added to older pages.

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
```diff
@@ -74,29 +80,20 @@ class ReportResourceTest : SqsIntegrationTestBase() {
...
       fun `access forbidden with wrong role`() {
-        webTestClient.get().uri("/incident-reports/${existingReport.id}")
+        webTestClient.get().uri(url)
           .headers(setAuthorisation(roles = listOf("ROLE_BANANAS")))
           .header("Content-Type", "application/json")
           .exchange()
           .expectStatus().isForbidden
       }
-
-      @Test
```
xoen, line 84: Is this test gone? I see there are new similar tests but not for the `GET /incident-reports/${existingReport.id}` endpoint? (I'm probably missing something)

ushkarev: it's doing the wrong thing. `MAINTAIN_INCIDENT_REPORTS` is _not_ needed. and `read` is the _right_ scope (tho the endpoint does not require any particular scope)

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
ushkarev: i updated my comment a bit late but we cannot sort the reports by id and have the order be stable since _these_ uuids are essentially random.

still, it’s up to the client to decide what field to order by 🤷

xoen: Oh they're UUID, so that's true! In that case sorting by date is more sensible.

Maybe sort by creation date would be better and there is less chance of newer reports getting in the "middle"?

ushkarev: >Maybe sort by creation date would be better and there is less chance of newer reports getting in the "middle"?

creation date is still a problem for reports that come from nomis:

>the reconciliation process should probably not use a date as it won't result in a stable order (they’re non-unique columns). actually, especially so with created date – we don’t migrate that info from NOMIS so it’s the moment of migration that gets stored.

Mjwillis: Yes don't use UUID for sorting. But we have the incident reporting number which is unique

PR #82, opened 2024-05-16
IR-236: Replace UUID generation with v7

ushkarev:
…to have report primary keys be (almost certainly) lexicographically sequential. This will enable sorting reports in the database using the `id` column and retaining a stable order.

Note that:
• the current `org.hibernate.id.UUIDGenerator` is deprecated anyway
• and the suggested replacement `org.hibernate.id.uuid.UuidGenerator` only supports UUID v1 and v4
• deployed databases will need to be cleared prior to merging

### DIFF
--- a/src/main/resources/db/migration/V1_1__initial_schema.sql
+++ b/src/main/resources/db/migration/V1_1__initial_schema.sql
@@ -21,7 +21,7 @@ create table event
 
 create table report
 (
-    id                     uuid       default gen_random_uuid() not null
+    id                     uuid                                 not null
         constraint report_pk primary key,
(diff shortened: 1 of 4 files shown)


src/main/resources/db/migration/V1_1__initial_schema.sql
```diff
@@ -21,7 +21,7 @@ create table event
 
 create table report
 (
-    id                     uuid       default gen_random_uuid() not null
+    id                     uuid                                 not null
```
ushkarev, line 24: ensures that generation happens in code, not in database

xoen (approved):

PR #86, opened 2024-05-20
IR-240: Improve filtering swagger docs

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResource.kt
```diff
@@ -90,6 +90,7 @@ class ReportResource(
       required = false,
       defaultValue = "null",
       example = "DPS",
+      implementation = InformationSource::class,
```
ushkarev, line 93: this uses the enum to list allowable values

xoen (approved):

PR #87, opened 2024-05-21
IR-239: Map new NOMIS sync fields

xoen:
Some additional metadata fields (created/updated timestamps, created/updated by, etc) have been added to the NOMIS sync requests: [See PR](https://github.com/ministryofjustice/hmpps-nomis-prisoner-api/pull/719).

This PR adds these fields to our NOMIS DTO classes and uses them when makes more sense.

- Report's `createdDate` is now set to whatever is in the NOMIS sync requests (instead of just using "now")
- same for `lastModifiedDate`/`lastModifiedBy` (although these are optional fields in the NOMIS sync request, so we fallback to `createDateTime`/`createdBy` when missing)
- these values are also used for the overarching `Event`
   
A nice side effect of this changes is that `toNewEntity()` doesn't need a dependency to `Clock` and also `updateWith()` doesn't need an `updatedBy` argument anymore (as this information is already in the upsert NOMIS sync Request)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/NomisReport.kt
```diff
@@ -34,6 +35,19 @@ data class NomisReport(
...
+  @Schema(description = "The username of the person who created the incident")
+  val createdBy: String,
+
+  @Schema(description = "The date and time the incident was last updated")
+  val lastModifiedDateTime: LocalDateTime? = createDateTime,
+  @Schema(description = "The username of the person who last updated the incident")
+  val lastModifiedBy: String? = createdBy,
+
+  @Schema(description = "The follow up date for the incident")
+  val followUpDate: LocalDate? = null,
```
ushkarev, line 49: should we leave a to-do message noting that we currently ignore this?

ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/NomisReport.kt
xoen: Not sure, this is not the only field in the NOMIS sync request we are ignoring...it may get a bit too noisy if we add a note on each fields 🤔

PR #93, opened 2024-05-22
IR-242: Add deletion endpoint

ushkarev:
NOMIS sync needs to be able to delete reports, but it makes sense to promote this facility to be general purpose. Even if the incident reporting user-facing service does not allow deleting reports, the incident report _api_ ought to provide the mechanism anyway (being a CRUD REST api).

Now:
• reports can be deleted by primary key
• write scope is needed, but currently, only the regular `MAINTAIN_INCIDENT_REPORTS` role is required
• deletion is logged and a domain event is raised
• orphaned events are deleted by default

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
@@ -67,7 +67,7 @@ class Report(
   var reportedBy: String,
   var reportedDate: LocalDateTime,
 
-  @ManyToOne(fetch = FetchType.LAZY, cascade = [CascadeType.ALL], optional = false)
+  @ManyToOne(fetch = FetchType.LAZY, cascade = [CascadeType.PERSIST, CascadeType.MERGE, CascadeType.REFRESH, CascadeType.DETACH], optional = false)
   val event: Event,
 
   @OneToMany(mappedBy = "report", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -67,8 +67,8 @@ class Report(
   var reportedBy: String,
   var reportedDate: LocalDateTime,
 
-  @ManyToOne(fetch = FetchType.LAZY, cascade = [CascadeType.ALL], optional = false)
-  val event: Event,
+  @ManyToOne(fetch = FetchType.LAZY, cascade = [CascadeType.PERSIST, CascadeType.MERGE, CascadeType.REFRESH, CascadeType.DETACH], optional = false)
```
ushkarev, line 70: cascading delete operations from report to event meant that removing a single report resulted in _all_ sibling reports and owning event also being removed

xoen: Why sibling reports would be deleted? That's odd 🤔

ushkarev: we definitely don't want them to be. but the cascade attribute used to include cascading deletion, so when a report was deleted, the parent event was deleted and so all other reports under that were also deleted.

ushkarev: now, deleting the event must be done actively rather than implicitly by cascade

PR #95, opened 2024-05-23
IR-246: Improve request payload validation

ushkarev:


src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/SyncServiceTest.kt
```diff
@@ -384,22 +383,4 @@ class SyncServiceTest {
     // verify telemetry not sent
     verify(telemetryClient, never()).trackEvent(anyOrNull(), anyOrNull(), anyOrNull())
   }
-
-  @Test
-  fun `cannot update report during initial migration`() {
```
ushkarev, line 389: this is covered by `NomisSyncResourceTest.SamplePayloads.cannot update a report during initial migration`.

this test cannot be performed here because the service is instantiated directly without spring’s proxying (and hence no input validation).

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/CreateReportRequest.kt
```diff
@@ -32,10 +32,16 @@ data class CreateReportRequest(
   val reportedDate: LocalDateTime,
   // TODO: there is not yet a way to add any more details to a report, question-response pairs, etc
 ) {
-  fun toNewEntity(incidentNumber: String, event: Event, createdBy: String, clock: Clock): Report {
+  fun validate() {
```
xoen, line 35: this should probably be private

xoen: hang on, I need to have a look at this better as I thought this was invoked directly from within `toNewEntity()` 🤔

xoen (dismissed):

xoen (approved):

PR #96, opened 2024-05-23
IR-245: Tidy up metadata field names

xoen:
Renamed many of the metadata fields (timestamps/users that did something) for consistency and reduce confusion (for example `changeDate`/`createdDate` are actually datetimes):
- use `At` suffix when the field is a datetime (e.g. `createdAt`, `modifiedAt`, etc...)
- use `By` suffix when the field is a username (e.g. `modifiedBy`)
- replaced `setOn`/`setBy` to `changedAt`/`changedBy` so that histories have consistent fields

### API breaking changes
As our DTO classes changes **this is a breaking change** as the response format changes.

I've already mentioned this to Syscon, so they're aware of this change, we don't have any other clients yet so this is a good time to tidy up.

### NOMIS sync DTOs
The NOMIS DTOs are **_not_** changing as this is the format of the requests we're receiving.

### DB schema / JPA
The JPA entities and the DB schema changed, I've made these changes directly into the initial migration which means DB will need to be reset.

⚠️ **DO NOT MERGE YET**. @ushkarev said to hold off merging this just yet.


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/Mapping.kt
```diff
@@ -20,12 +20,12 @@ fun NomisReport.toNewEntity(): Report {
     title = title ?: "NO DETAILS GIVEN",
     description = description ?: "NO DETAILS GIVEN",
     reportedBy = reportingStaff.username,
-    reportedDate = reportedDateTime,
+    reportedAt = reportedDateTime,
```
xoen, line 23: note that the field names on the right are not changed as these are coming from NOMIS and those DTOs are not changing.

ushkarev (approved):

PR #100, opened 2024-05-28
IR-248 Add EntityOpen annotation and change fetch type to LAZY

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -67,38 +69,38 @@ class Report(
   var reportedBy: String,
   var reportedAt: LocalDateTime,
 
-  @ManyToOne(fetch = FetchType.EAGER, cascade = [CascadeType.PERSIST, CascadeType.MERGE, CascadeType.REFRESH, CascadeType.DETACH], optional = false)
+  @ManyToOne(fetch = FetchType.LAZY, cascade = [CascadeType.PERSIST, CascadeType.MERGE, CascadeType.REFRESH, CascadeType.DETACH], optional = false)
```
Mjwillis, line 72: This could be a real performance problem

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/HistoricalQuestion.kt
```diff
@@ -26,7 +28,7 @@ class HistoricalQuestion(
 
   val additionalInformation: String? = null,
 
-  @OneToMany(mappedBy = "historicalQuestion", fetch = FetchType.EAGER, cascade = [CascadeType.ALL], orphanRemoval = true)
+  @OneToMany(mappedBy = "historicalQuestion", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
```
ushkarev, line 31: actually, it's the other way round! we used to have the lazy type annotation and that spawned a HUGE number of sql statements. changing to eager helped… ref #97

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
xoen: Don't we use most (if not all) the associated records of a report anyway? (e.g. when serialising to the `Report` DTO response class).

That means loading these associated records is actually slower as they would be loaded anyway lazily via additional SQL queries (the so called "N+1 queries problem")

Mjwillis: Trust me on this one.  Yes its true when you are loading the whole report but the problem is that it will ALWAYS load this data on ANY query you run.  So for you full DTO convert we can perform a batch fetch in the query - or use a `@NamedEntityGraph`

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/HistoricalQuestion.kt
Mjwillis: For one class with only one 1:M it might be ok, but its always better to explicity fetch not by default.

ushkarev:
This PR is being semi-abandoned for now in favour of #115 which will simply _not_ return all the related entities when listing many reports (which is never likely to be fast no matter what JOINs or optimisations are made).

A separate endeavour exists to replace the “eager” annotations with named entity graphs (see entity-graph* branches) – this works well for the single report endpoints, but is incomplete for the many-reports endpoint.

PR #112, opened 2024-06-05
IR-266 Return only ID of report in sync

Mjwillis:
The returned result now contains report ID, instead of the full report.

NOMIS sync doesn't need the full payload of the report returned just the response to indicate what has been updated.


src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/NomisSyncResourceTest.kt
```diff
@@ -409,237 +409,6 @@ class NomisSyncResourceTest : SqsIntegrationTestBase() {
           .bodyValue(jsonString(updatedSyncRequest))
           .exchange()
           .expectStatus().isCreated
-          .expectBody().json(
```
ushkarev, line 412: maybe these body assertions can become something like
```kotlin
.expectBody().consumeWith {
  val reportId = objectMapper.readValue(it.responseBody, UUID::class.java)
  assertThat(reportId).isNotNull()
}
```
so that we actually ensure a uuid is returned?

ushkarev: also, the problem with removing the JSON response assertions is that now we have nothing assuring the right stuff got saved in the db

Mjwillis: Yes I realise that - feel like that test should be elsewhere

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/EventPublishAndAuditService.kt
```diff
@@ -19,26 +19,25 @@ class EventPublishAndAuditService(
...
         id = reportId.toString(),
         auditData = it,
-        source = source,
       )
     }
   }
 
-  private fun publishEvent(
-    event: ReportDomainEventType,
+  private fun domainEvent(
```
xoen, line 33: I think we should retain the "publish" in this function name, maybe rename to `publishDomainEvent()`, but I like the "publish" in it as it makes it clear what this is doing, especially in the client code, e.g.

```
domainEvent(eventType = eventType, reportId = reportId, source = source)
```

seeing this doesn't tell me much, it almost seems like it's constructing an object but not using it.

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/EventPublishAndAuditService.kt
```diff
@@ -47,11 +46,10 @@ class EventPublishAndAuditService(
     )
   }
 
-  fun auditEvent(
+  private fun auditEvent(
```
xoen, line 49: possibly rename this to be `sendAuditEvent()` as well

xoen (approved):

PR #115, opened 2024-06-06
IR-265: Return only key report info when listing many reports

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResource.kt
```diff
@@ -159,7 +160,7 @@ class ReportResource(
     @ParameterObject
     @PageableDefault(page = 0, size = 20, sort = ["incidentDateAndTime"], direction = Sort.Direction.DESC)
     pageable: Pageable,
-  ): SimplePage<ReportDto> {
+  ): SimplePage<ReportBasic> {
```
xoen, line 163: Do you think it's worth changing the name of the Resource method to `getBasicReports()`? (and others accordingly as well?) - may help when navigating/orientating these big resources files.

xoen: Similarly, the method name in the Service is now a bit ambiguous: `getReports()` ...are we getting back the basic ones or the ones with details?

xoen: Other methods in the service may do different things, so there may be a consistency problem which may be surprising in the future.

For example, `getReports()` returns `ReportBasic` but `getReportById()` returns `ReportWithDetails?`.

I don't know if you like the idea of having more explicit method names  🤔

xoen (approved):

PR #116, opened 2024-06-10
IR-68: Have database sequences be owned by tables

ushkarev:
…so that dropping the tables or columns cascades dropping the sequence


ushkarev:
this might be a little pointless, i'm not wedded to it :)

xoen (approved):
This means we don't have to manually drop the sequences after deleting the tables when clearing the DB for migration testing right? Sounds like a good idea 👍

PR #117, opened 2024-06-10
IR-68: Turn time zone id into a separate, shared, bean

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/config/TimeZoneConfiguration.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import java.time.ZoneId
+
+@Configuration
+class TimeZoneConfiguration(
+  @Value("\${spring.jackson.time-zone}") private val timeZone: String,
+) {
+  @Bean
+  fun timeZone(): ZoneId = ZoneId.of(timeZone)
```
xoen, line 13: Maybe call the bean `TimeZoneId`?

xoen (dismissed):
🫘

PR #122, opened 2024-06-13
IR-277: Update report api

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResource.kt
```diff
@@ -357,11 +357,6 @@ class ReportResource(
         description = "Data not found",
         content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
       ),
-      ApiResponse(
-        responseCode = "409",
-        description = "Incident report already exists",
```
ushkarev, line 362: the request to create a new draft report does not give the caller an option to provide any ids so this cannot happen

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/CreateReportRequest.kt
```diff
@@ -27,9 +30,10 @@ data class CreateReportRequest(
   @Schema(description = "Which existing event to link to", required = false, defaultValue = "null")
   val linkedEventId: String? = null,
   @Schema(description = "Username of person who created the incident report", required = true)
+  @field:Size(min = 3, max = 120)
   val reportedBy: String,
   @Schema(description = "When the incident report was created", required = true, example = "2024-04-29T12:34:56.789012")
-  val reportedDate: LocalDateTime,
```
ushkarev, line 32: might as well match the jpa entity and output dtos

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/UpdateReportRequest.kt
```diff
@@ -24,6 +24,9 @@ data class UpdateReportRequest(
   val reportedBy: String? = null,
   @Schema(description = "When the incident report was created", required = false, defaultValue = "null", example = "2024-04-29T12:34:56.789012")
   val reportedAt: LocalDateTime? = null,
+
+  @Schema(description = "Whether the parent event should also be updated", required = false, defaultValue = "false", example = "true")
+  val updateEvent: Boolean = false,
```
ushkarev, line 29: report deletion already has an option to propagate to the parent event, so it makes sense to allow propagating field updates. there are no other fields in the event than those already listed here – we might no need any special event-editing endpoints!

of course, the parent event is very likely to change shape and be renamed in future

PR #124, opened 2024-06-14
IR-280: CRUD API endpoints for related objects

ushkarev:
Specifically:
• Staff involvement
• Prisoner involvement
• Location
• Evidence
• Correction requests

History, historical questions, historical responses and status history are read-only.

Questions and responses will be dealt with in a separate PR.


src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
```diff
@@ -1550,16 +1551,24 @@ class ReportResourceTest : SqsIntegrationTestBase() {
...
-            .exchange()
-            .expectStatus().isBadRequest
-
-          assertNoDomainMessagesSent()
+        @DisplayName("cannot add invalid object to a report")
+        @TestFactory
+        fun `cannot add invalid object to a report`(): List<DynamicTest> {
+          val requests = invalidRequests.toMutableMap()
+          requests["empty request"] = "/related-objects/empty-object.json"
+          requests["invalid shape"] = "/related-objects/empty-array.json"
```
xoen: these two could be inlined maybe as they're not specific of any of these resources, it may make the test a bit simpler.

ushkarev: not cleanly cos the test tries to load the file at that path. it'd have to have a way to know whether it should 1) load a file or 2) use a string directly

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
```diff
@@ -1145,12 +1130,7 @@ class ReportResourceTest : SqsIntegrationTestBase() {
             true,
           )
 
-        getDomainEvents(1).let {
-          assertThat(it.map { message -> message.eventType to message.additionalInformation?.source })
-            .containsExactlyInAnyOrder(
-              "incident.report.amended" to InformationSource.DPS,
-            )
-        }
+        assertThatDomainEventWasSent("incident.report.amended", "IR-0000000001124143")
```
xoen, line 1133: Too verbose to use the enum maybe?

ushkarev: the value inside the event that's sent is a string so the enum would need to be converted to a string… which feels less accurate cos you then never truly test the actual value…?

ushkarev: just had a look and the enum variant name is equally long :)

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
xoen: aren't the 2 "requests in this test just strings tho 🤔?

one [of the tests was using string literals](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/124/files/7beec863dd26c39df358e91f1170b1f19c020042..5274daea0455abdcd8872441b3cefee6ff437ee6#diff-d6a0e69b7f1bf04510939e174746671e9585d2c1892aa614a0ca11211c1ac26dL1682) before no?
 
I don't mind if you think this is better or it's not possible to simplify, but I think there are a lot of fixtures files now ☺️

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
xoen: Yeah that is long, but you could call `.value` on the enum AFAICT. I'd use the constant but I don't mind using these strings as I imagine they're pretty much stable :)

xoen (approved):
I'm not super keen on the use of indexes for update/deletion.

But if we think the risk of race condition is minimal/very low and we think this makes for a better API I don't mind keeping this, we can change if necessary I guess 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
ushkarev: actually i worked a way out. i don't think it makes things simpler really, but it does get rid of these stupid empty-ish json files

ushkarev:
> I'm not super keen on the use of indexes for update/deletion.

i don't love the idea of proliferating most of our objects with ids, but it's certainly an option! these lists of “related objects” have a stable order and i suspect that concurrent modification is quite unlikely.

setting aside that the responses will be less clean, it won't be trivial to move to ids because the test data will have unstable ids so looking things up and making assertions will be more complicated.

tho i'm not set against it :) i suggest we wait until we have clearer design plans…?

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
ushkarev: while not terrible, i see the downside of something like this
```kotlin
assertThatDomainEventWasSent(INCIDENT_REPORT_AMENDED, "IR-0000000001124143")
```
as tests not picking up as many typos should they happen

it's for a similar reason that i like that we have a fixed clock so that all json responses can use an exact string representation rather than `$now`

xoen (approved):

PR #126, opened 2024-06-18
IR-281 & IR-282: Allow changing a report’s status or type

ushkarev:
Changing a report’s status or type has a side effect of preserving the history, so there are now separate endpoints from the generic PATCH report method.

Changing a type to an inactive one is not allowed.

We still do not know what the lifecycle of a report is so there is no restriction on what transitions are allowed for status. Should the api permit any status change? Should it instead check the user’s role and disallow some? For instance, perhaps only QAs or central support can re-open a report once it has been closed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/utils/MaybeChanged.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/utils/MaybeChanged.kt
@@ -1,0 +1,24 @@
+package uk.gov.justice.digital.hmpps.incidentreporting.dto.utils
+
+/**
+ * Represents a value that might have been changed,
+ * allowing a block to be called only if it had been changed.
+ */
+sealed interface MaybeChanged<T> {
+  val value: T
+
+  fun alsoIfChanged(block: (T) -> Unit): MaybeChanged<T>
+
+  data class Unchanged<T>(override val value: T) : MaybeChanged<T> {
+    override fun alsoIfChanged(block: (T) -> Unit): Unchanged<T> {
+      return this
+    }
+  }
+
+  data class Changed<T>(override val value: T) : MaybeChanged<T> {
+    override fun alsoIfChanged(block: (T) -> Unit): Changed<T> {
+      block(value)
+      return this
+    }
+  }
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
@@ -18,6 +18,7 @@ import uk.gov.justice.digital.hmpps.incidentreporting.constants.Type
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.ReportBasic
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.ReportWithDetails
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.request.ChangeStatusRequest
+import uk.gov.justice.digital.hmpps.incidentreporting.dto.request.ChangeTypeRequest
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.request.CreateReportRequest
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.request.UpdateReportRequest
 import uk.gov.justice.digital.hmpps.incidentreporting.jpa.repository.EventRepository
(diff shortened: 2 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
```diff
@@ -182,4 +185,64 @@ class ReportService(
       }
     }.getOrNull()
   }
+
+  @Transactional
+  fun changeReportStatus(id: UUID, changeStatusRequest: ChangeStatusRequest): Pair<ReportWithDetails, Boolean>? {
```
ushkarev, line 190: it's a bit unpleasant returning a “pair” or values. the report might or might not have been updated and this is reflected by the boolean flag…

we can make something a bit like a tagged union?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
```diff
@@ -182,4 +184,34 @@ class ReportService(
...
+          newStatus = changeStatusRequest.newStatus,
+          changedAt = now,
+          changedBy = user,
+        )
+        it.modifiedAt = now
+        it.modifiedBy = user
+      }
+
+      it.toDtoWithDetails().apply {
+        if (changed) {
```
xoen: Couldn't you move this inside the `if (changed)` block above and avoid this other `if`?

It's not much about the if itself but I think it makes sense to keep the two inside the same block as they're only happening together anyway (same condition)

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
xoen: This returns the `ReportWithDetails` when the report is changing.

So I think returning `ReportWithDetails?` maybe be simpler and good enough.

e.g. returns a `ReportWithDetails` if the status changed, returns `null` when the status didn't change.

xoen: I see, it's because you want this to also check if the report exists?

The client code could either pass the Report or you could just throw a "Report not found exception" in here (which I think it makes sense as this is asked to update a report and if it doesn't exist it's an invalid operation)

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
```diff
@@ -1444,6 +1444,377 @@ class ReportResourceTest : SqsIntegrationTestBase() {
...
+          .headers(setAuthorisation(roles = listOf("ROLE_MAINTAIN_INCIDENT_REPORTS"), scopes = listOf("write")))
+          .header("Content-Type", "application/json")
+          .bodyValue(
+            // language=json
+            """
+              {"newType": "FINDS"}
+            """,
+          )
+          .exchange()
+          .expectStatus().isOk
```
xoen, line 1575: Should this be allowed given type is not changing? 🤔

ushkarev: i think it should 404 if not found, but it's better to be idempotent in REST apis where possible. so if you set type to DAMAGE twice the second time should be ok but make no changes

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/utils/MaybeChanged.kt
```diff
@@ -0,0 +1,24 @@
...
+ * allowing a block to be called only if it had been changed.
+ */
+sealed interface MaybeChanged<T> {
+  val value: T
+
+  fun alsoIfChanged(block: (T) -> Unit): MaybeChanged<T>
+
+  data class Unchanged<T>(override val value: T) : MaybeChanged<T> {
+    override fun alsoIfChanged(block: (T) -> Unit): Unchanged<T> {
+      return this
```
xoen, line 14: Isn't it worth calling the block anyway? Maybe the client wants to do something on the unchanged thing? 🤔

xoen: I see, this method is just here so that client can call this regardless of the value but it would only invoke the block if stuff actually changed 👌

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
```diff
@@ -187,62 +188,64 @@ class ReportService(
...
-          log.info("Changed incident report status to ${changeStatusRequest.newStatus} for number=$incidentNumber ID=$id")
-          telemetryClient.trackEvent(
-            "Changed incident report status",
-            this,
-          )
-        }
-      } to changed
-    }.getOrNull()
+      maybeChangedReport.alsoIfChanged { reportWithDetails ->
+        log.info("Changed incident report status to ${changeStatusRequest.newStatus} for number=${reportWithDetails.incidentNumber} ID=$id")
```
xoen, line 212: I still think it could be moved in the if block above :)

xoen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
ushkarev: the event publishing thing isn't available in the service and i'm not sure i can wire it up properly :(

it's weird that the other service methods already had logging and telemetry in the service itself but event publishing outside it in the resource

PR #128, opened 2024-06-20
IR-292: CRUD API endpoints for questions and responses

ushkarev:
- Questions are always listed in order of insertion
    - …their responses are always listed in order of insertion
- Question and response IDs are not exposed
- A question can be added to a report if it has at least one response
- The last question (and all of its responses) can be deleted if it exists
- Because subsequent questions often depend on prior responses, you cannot modify any existing questions nor delete any but the last one
    - Should there be a need to modify something in the middle, it and all subsequent questions should be deleted and recreated as needed
---
questions-api

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/AddQuestionWithResponses.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/AddQuestionWithResponses.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.incidentreporting.dto.request
 
 import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.Valid
 import jakarta.validation.constraints.Size
 import java.time.LocalDateTime
 
@@ -10,8 +11,11 @@ data class AddQuestionWithResponses(
   @field:Size(max = 60)
   val code: String,
   @Schema(description = "The question", required = true)
+  @field:Size(min = 1)
   val question: String,
   @Schema(description = "The responses to this question", required = true)
+  @field:Valid
+  @field:Size(min = 1)
   val responses: List<AddQuestionResponse>,
   @Schema(description = "Optional additional information", required = false, defaultValue = "null")
   val additionalInformation: String? = null,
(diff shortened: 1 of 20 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/AddQuestionWithResponses.kt
```diff
@@ -0,0 +1,30 @@
...
+)
+
+data class AddQuestionResponse(
+  @Schema(description = "The response", required = true)
+  val response: String,
+  @Schema(description = "Username of person who responded to the question", required = true)
+  @field:Size(min = 3, max = 120)
+  val recordedBy: String,
+  @Schema(description = "When the response was made", required = true, example = "2024-04-29T12:34:56.789012")
+  val recordedAt: LocalDateTime,
```
xoen, line 32: Any reason to ask the clients to pass this?

I'm assuming this API is used to submit questions so in theory we can set this to "now" when we persist them no?

At the very least we should make this optional and default to now if you think it's still valuable to allow clients to override this.

ushkarev: it's a good question. i sorta propagated the pattern from the original `CreateReportRequest.reportedBy` without thinking too much about it. which do you think is better…
1) remove recordedBy and recordedAt here and use `authenticationFacade.getUserOrSystemInContext()` and `LocalDateTime.now(clock)` – because those 2 are used for the parent report's modifiedBy and modifiedAt
2) as you suggest, make them optional but _also_ propagate them to the report's modifiedBy and modifiedAt

ushkarev: i dunno which is better :/

ushkarev: nomis screenshots imply that it might be possible for different responses to 1 question to have been made by different users…

xoen: To recap our chat: Consistency aside (which I agree is important for an API), I think my preference would be to not have these fields and set them automatically to "current real user"/"now".

When/_if_ the need to override these good default ever arise we can always update the API to have optional fields which still default to these values (and this would be a non-breaking change for existing clients).

xoen (approved):

PR #129, opened 2024-06-21
IRS-291: Domain events include an hint of what part of the report changed

xoen:
When a Report is updated the published `incident.report.amended` has a `whatChanged` field in the `additionalInformation`.

This field is an enum that the consumers of these domain events can use to have a better idea of what part of the report changed, e.g. prisoners involved, staff involved, etc...

If the endpoint to update only the basic information ([`PATCH /incident-reports/{id}`](https://incident-reporting-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/Incident%20reports/updateReport)) was used this field would be set to `BASIC_REPORT` to indicate that only the Report key properties may have changed.

The sync upsert endpoint is a bit special in the sense that the client pass the whole Report's blob.
This means that any part of the Report could change (and potentially different parts at the same time).
In this case the `whatChanged` field would be set to `ANYTHING` to indicate that any part of the report may have changed (this is vague but still less ambiguous than setting the field to `null`)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportRelatedObjectsResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportRelatedObjectsResource.kt
@@ -13,6 +13,7 @@ import uk.gov.justice.digital.hmpps.incidentreporting.jpa.Report
 import uk.gov.justice.digital.hmpps.incidentreporting.jpa.repository.ReportRepository
 import uk.gov.justice.digital.hmpps.incidentreporting.service.ReportDomainEventType
 import uk.gov.justice.digital.hmpps.incidentreporting.service.ReportService.Companion.log
+import uk.gov.justice.digital.hmpps.incidentreporting.service.WhatChanged
 import java.time.Clock
 import java.time.LocalDateTime
 import java.util.UUID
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportRelatedObjectsResource.kt
```diff
@@ -41,17 +42,17 @@ abstract class ReportRelatedObjectsResource<ResponseDto, AddRequest, UpdateReque
...
-  protected fun <T> (UUID).updateReportOrThrowNotFound(changeMessage: String, block: (Report) -> T): T {
+  protected fun <T> (UUID).updateReportOrThrowNotFound(changeMessage: String, whatChanged: WhatChanged? = null, block: (Report) -> T): T {
     val report = findReportOrThrowNotFound()
     return block(report).also {
       report.modifiedAt = LocalDateTime.now(clock)
       report.modifiedBy = authenticationFacade.getUserOrSystemInContext()
 
       val basicReport = report.toDtoBasic()
       eventPublishAndAudit(
-        // TODO: should different related object actions raise different events?
```
xoen, line 52: @ushkarev I think this should address this TODO so [I've removed it](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/129/commits/d952ee3eaffe2a06717488e7b2eb1236e35c8eee).

I believe this is a good middle ground and better than having a lot of possible event types for each part of the Report that may or may have not changed.

The result should be similar: Consumer knows that a Report has been updated and what part (roughly) changed - but it's less unwieldy than having another 10 or so domain event types ☺️

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/SnsService.kt
```diff
@@ -65,10 +65,26 @@ class SnsService(
...
+  TYPE, // change to the report type
+  STATUS, // change to the report status
+  PRISONERS_INVOLVED,
+  STAFF_INVOLVED,
+  LOCATIONS,
+  EVIDENCE,
+  CORRECTION_REQUESTS,
+
+  // TODO: Just questions? Questions/answers? Endpoint not there yet
+  QUESTIONS,
```
xoen, line 80: @ushkarev FYI: I've just added a `QUESTION` discriminant to this enum but I don't know how the endpoint(s) will be shaped so not sure if this is enough or needs tweaking.

ushkarev (approved):

PR #132, opened 2024-06-25
IR-304: Take actor username from authentication token rather than explicit parameters

ushkarev:
All modifying-type api endpoints record the time the change happens and the actor, the username of who did it. Some endpoints allow these to be overridden in the JSON payload, but most do not. The suggestion (for now) is to remove the ability to override these and only use the username from the oauth token and the true “now”.

These payload fields will be removed:
- `AddQuestionResponse.recordedBy` and `AddQuestionResponse.recordedAt`
- `AddCorrectionRequest.correctionRequestedBy` and `AddCorrectionRequest.correctionRequestedAt`
- `UpdateCorrectionRequest.correctionRequestedBy` and `UpdateCorrectionRequest.correctionRequestedAt`
- `CreateReportRequest.reportedBy` and `CreateReportRequest.reportedAt` †
- `UpdateReportRequest.reportedBy` and `UpdateReportRequest.reportedAt` †

† The last 2 bullets might be slightly contentious because a Report entity has a createdBy and reportedBy field; the idea is to give them the same value when called with new apis, but they will be different for data synchronised from NOMIS.

The assumption is that all apis will use system tokens with the username of the actor – this should work fine for UI apps calling this api directly, but it could be harder for other apis to call IRS endpoints when there is not a clear actor. Eg. for some kind of scheduled tasks? It’s not very likely to be a big problem.

If it proves to be necessary, this work could be reversed, but then we ought to add the actor and time to all other endpoint payloads.

---

Additionally, empty update payloads should not mark the report as updated or raise domain events since nothing actually changed.


xoen (approved):
Simpler ❤️

PR #136, opened 2024-07-08
IR-68: Sync changes from templates

ushkarev:
Ref:
- [hmpps-ip-allowlists#21](https://github.com/ministryofjustice/hmpps-ip-allowlists/pull/21)
- [hmpps-template-kotlin#215](https://github.com/ministryofjustice/hmpps-template-kotlin/pull/215)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportCorrectionRequestResource.kt
```diff
@@ -115,7 +116,7 @@ class ReportCorrectionRequestResource : ReportRelatedObjectsResource<CorrectionR
         report.addCorrectionRequest(
           reason = reason,
           descriptionOfChange = descriptionOfChange,
-          correctionRequestedBy = authenticationFacade.getUserOrSystemInContext(),
+          correctionRequestedBy = authenticationHolder.username ?: SYSTEM_USERNAME,
```
petergphillips, line 119: I think we're trying in future to move to explicitly passing through users rather than putting them in the auth token.  That said I haven't seen this pattern before, but wondering if it would be a good idea for the holder to be initialised with a default username in `application.yaml` e.g. `hmpps.auth.default-username` or similar.  Would then have a separate method that would return a `String` instead of nullable `String?` e.g. `holder.usernameOrDefault`

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/utils/UserMdcFilter.kt
```diff
@@ -9,19 +10,21 @@ import org.slf4j.MDC
...
 import org.springframework.core.annotation.Order
 import org.springframework.stereotype.Component
-import uk.gov.justice.digital.hmpps.incidentreporting.config.AuthenticationFacade
+import uk.gov.justice.hmpps.kotlin.auth.HmppsAuthenticationHolder
 import java.io.IOException
 
 @Component
 @Order(1)
-class UserMdcFilter @Autowired constructor(private val authenticationFacade: AuthenticationFacade) : Filter {
+class UserMdcFilter @Autowired constructor(
```
petergphillips, line 18: I've opted to remove this filter in some of our classes, since it is only for `AppTraces` logging, `AppRequests` and `AppEvents` uses the [ClientTrackingConfiguration](https://github.com/ministryofjustice/hmpps-incident-reporting-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/config/ClientTrackingConfiguration.kt#L36) instead.

petergphillips: I've got a task to add the `ClientTrackingConfiguration` to the kotlin lib as well, but need to spend some time to work out if I can use the `HmppsAuthenticationHolder` in it as all existing implementations

ushkarev: i saw that you removed it, but didn't know what it does in the first place so was more cautious :p

on that note, i recall that you also removed `JwtAuthInterceptor`, but kept `UserContext` and `UserContextFilter` – should they all be dropped? it looked like the user context stuff was only referred to inside the auth interceptor. (again i have no idea what they are _supposed_ to do so couldn't check!)

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportCorrectionRequestResource.kt
ushkarev: i actually feel that the fallback username is a bit of a dirty hack. i've simply kept the logic the same as much as possible for the purpose of this pr. but in my opinion, for the use cases in this repo, we should _require_ the username if we actually expect it.

petergphillips: agreed, especially as you'd never notice if a client no longer passes through the `username` in the token.  Hence my comment that it should be passed through explicitly instead.

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/utils/UserMdcFilter.kt
petergphillips: I think in your case they can all be dropped.  Some uses of the `UserContext` are to grab the authentication out of the request to then set for subsequent calls e.g. https://github.com/ministryofjustice/cmd-api/blob/de106d425ed822c6c6fbc1fcd4a5885f1170e2c9/src/main/java/uk/gov/justice/digital/hmpps/cmd/api/config/WebClientConfiguration.kt#L82, but can't see that being used here

Mjwillis (approved):
LGTM

PR #140, opened 2024-07-17
IR-372: Add read-only events endpoints

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/EventResource.kt
```diff
@@ -0,0 +1,151 @@
...
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.incidentreporting.dto.Event
+import uk.gov.justice.digital.hmpps.incidentreporting.dto.response.SimplePage
+import uk.gov.justice.digital.hmpps.incidentreporting.dto.response.toSimplePage
+import uk.gov.justice.digital.hmpps.incidentreporting.service.EventService
+import java.util.UUID
+
+@RestController
+@Validated
+@RequestMapping("/incident-events", produces = [MediaType.APPLICATION_JSON_VALUE])
```
xoen, line 33: As mentioned offline (so just to recap), I am a bit wary of starting to use this "event" word in a public facing API.
I'm not sure this is the domain word for this or if it is actually "incident" (which I think it's the correct one, but maybe I'm wrong).

In theory it doesn't matter much, however "event" is quite generic and it is also used in other contexts (e.g. "domain events" in queues) so it makes things a bit more annoying to search.

Just mentioning this here again, nothing to do yet...unless as you mentioned we want to discuss this with other devs and possibly Katherine and see if they think "incident" is a better name for this concept (I think it is).

We may even potentially rename and display in different ways in the UI 🤷

ushkarev: to get down the reasoning we talked about:
1) `Event` already existed in the code; originally as `IncidentEvent`
2) it’s not a _terrible_ description of something that happened which involves collecting several reports together
2a) even though it is perhaps too generic and `Incident` would be clearer
3) there are no consumers of this api (at least parts related to whatever we call this entity)
4) the team is yet to decide on generally-accepted terms for the various parts
4a) it is even possible, though unlikely, that this entity will be removed

essentially: we can rename it if/when we need to :)

xoen (approved):

PR #146, opened 2024-08-01
IR-395: React to HMPPS prisoner merge domain event

ushkarev:
…by replacing given prisoner number in all involvement entities.

Open question remains: if reports and events will have auto-generated titles, and if prisoner numbers are included, we should rename these too

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
@@ -26,6 +26,7 @@ import uk.gov.justice.digital.hmpps.incidentreporting.dto.request.CreateReportRe
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.request.UpdateReportRequest
 import uk.gov.justice.digital.hmpps.incidentreporting.dto.utils.MaybeChanged
 import uk.gov.justice.digital.hmpps.incidentreporting.jpa.repository.EventRepository
+import uk.gov.justice.digital.hmpps.incidentreporting.jpa.repository.PrisonerInvolvementRepository
 import uk.gov.justice.digital.hmpps.incidentreporting.jpa.repository.ReportRepository
 import uk.gov.justice.digital.hmpps.incidentreporting.jpa.repository.generateEventReference
 import uk.gov.justice.digital.hmpps.incidentreporting.jpa.repository.generateReportReference
@@ -50,6 +51,7 @@ import kotlin.jvm.optionals.getOrNull
 class ReportService(
   private val reportRepository: ReportRepository,
   private val eventRepository: EventRepository,
+  private val prisonerInvolvementRepository: PrisonerInvolvementRepository,
   private val telemetryClient: TelemetryClient,
   private val authenticationHolder: HmppsAuthenticationHolder,
   private val clock: Clock,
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
```diff
@@ -323,4 +325,35 @@ class ReportService(
...
+        }
+    }
+      .values
+      .map { it.toDtoBasic() }
+      .sortedBy { it.id }
+      .also {
+        if (it.isNotEmpty()) {
+          log.info("Prisoner $removedPrisonerNumber merged into $prisonerNumber")
+          // TODO: track with telemetryClient for every report?
+          // TODO: should a domain event be raised? probably not
```
Mjwillis, line 355: No need to raise event

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
```diff
@@ -323,4 +325,35 @@ class ReportService(
...
+          }
+        }
+    }
+      .values
+      .map { it.toDtoBasic() }
+      .sortedBy { it.id }
+      .also {
+        if (it.isNotEmpty()) {
+          log.info("Prisoner $removedPrisonerNumber merged into $prisonerNumber")
+          // TODO: track with telemetryClient for every report?
```
Mjwillis, line 354: Think a telemetry event would be useful

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
```diff
@@ -323,4 +325,35 @@ class ReportService(
...
     }
   }
+
+  /**
+   * Replaces one prisoner number in all report-related entities with another.
+   * NB: free text fields are _not_ changed
+   */
+  @Transactional
+  fun replacePrisonerNumber(removedPrisonerNumber: String, prisonerNumber: String): List<ReportBasic> {
+    // TODO: maybe free text fields should be replaced too? especially if we end up generating report titles automatically
```
Mjwillis, line 335: Interesting, would the title not be auto generated ?

Mjwillis (dismissed):
Couple of comments but LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
ushkarev: that's the point – so far the plan seems to be to not give users a chance to enter a title and there was talk of auto-generating one based on type and prisoner involvement etc. *if* that's what happens, this method ought to replace references in the title too.

i think at this stage it's best to leave the TODO msg since we don't know what titles will be like…

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
ushkarev: ok. will add…

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
xoen: (Unless titles have names instead of prisoner numbers...but then again, TBD)

Mjwillis (approved):

PR #148, opened 2024-08-05
IR-395: Fix unpacking prisoner merge event fields

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/PrisonOffenderEventListener.kt
```diff
@@ -29,8 +29,10 @@ class PrisonOffenderEventListener(
     when (eventType) {
       PRISONER_MERGE_EVENT_TYPE -> {
         val mergeEvent = mapper.readValue(message, HMPPSMergeDomainEvent::class.java)
-        val (removedPrisonerNumber, prisonerNumber) = mergeEvent.additionalInformation
```
ushkarev, line 32: these 2 fields are defined in the opposite order in the data class so are unpacked incorrectly here!

xoen (approved):

PR #151, opened 2024-08-09
IR-401: Add date field to responses

ushkarev:
In NOMIS, some responses ask the user to enter a date alongside. This was missed and previously only the comment / additional information field was being handled.

In future, the DPS version of incident reporting may or may not ask users to enter dates, but historic information needs to still be preserved.

*NB: Will require databases to be trashed when deploying*


xoen (approved):
LGTM

PR #158, opened 2024-08-15
IR-405: Add new `DRONE2` report type

ushkarev:
NOMIS has now got 3 drone sightings types: DRONE (inactive and used in the past), DRONE1 (active and in use) and DRONE2 (active and in use)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/Type.kt
```diff
@@ -93,6 +101,7 @@ enum class Type(
       "CON_INDISC" -> OLD_CONCERTED_INDISCIPLINE
       "DISORDER" -> OLD_DISORDER
       "DRONE" -> OLD_DRONE_SIGHTING
+      "DRONE1" -> OLD_DRONE_SIGHTING1
```
xoen, line 104: Didn't you say `DRONE1` is still active tho? 🤔

xoen (dismissed):

PR #164, opened 2024-08-23
IR-381: Remove IE-/IR- prefix from event/report references

xoen:
The references are still a string but it doesn't have the `IE-`/`IR-` prefixes and also lost the leading 0s (as these would likely be ignored when searching and they'd lead to confusion when no event/report would match).

Bearing in mind our sequences start at 10,000,000 which means new DPS events/reports would have a reference which is at least `'10000000'` (while old NOMIS references would be lower as expected)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
@@ -69,7 +69,6 @@ class Report(
   /**
    * Human-readable reference. Previously known as ”incident number” in NOMIS.
    * A number when sourced from NOMIS.
-   * Prefixed with “IR-” when sourced from DPS.
    */
   @Column(nullable = false, unique = true, length = 25)
   val reportReference: String,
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -69,7 +69,6 @@ class Report(
   /**
    * Human-readable reference. Previously known as ”incident number” in NOMIS.
    * A number when sourced from NOMIS.
```
ushkarev, line 71: probs “A number when sourced from NOMIS _or DPS_.” instead? or can just be dropped?

ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
xoen: 👍 

https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/164/commits/35d4dcea19de3d3c9de1a43a5b1eacf2de0c45be

ushkarev (approved):

PR #168, opened 2024-08-28
IR-425: Lookup reports given staff or prisoner involvement

ushkarev:


ushkarev:
replaced by better implementation in #169

PR #171, opened 2024-08-28
Label types with OpenAPI attributes

ushkarev:
1) they are either read-only (used only for responses) or write-only (used only for requests)
2) some fields are nullable
3) for write-only request types, some fields are required and some are optional
4) stop using the deprecated `required` attribute

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/HistoricalQuestion.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/HistoricalQuestion.kt
@@ -3,15 +3,15 @@ package uk.gov.justice.digital.hmpps.incidentreporting.dto
 import com.fasterxml.jackson.annotation.JsonInclude
 import io.swagger.v3.oas.annotations.media.Schema
 
-@Schema(description = "Previous question with responses making up a previous version of an incident report")
+@Schema(description = "Previous question with responses making up a previous version of an incident report", accessMode = Schema.AccessMode.READ_ONLY)
 @JsonInclude(JsonInclude.Include.ALWAYS)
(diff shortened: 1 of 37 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/HistoricalQuestion.kt
```diff
@@ -3,15 +3,15 @@ package uk.gov.justice.digital.hmpps.incidentreporting.dto
...
+  @Schema(description = "The question code")
   val code: String,
-  @Schema(description = "The question", required = true)
+  @Schema(description = "The question")
   val question: String,
-  @Schema(description = "The responses to this question", required = true)
+  @Schema(description = "The responses to this question")
   val responses: List<HistoricalResponse> = emptyList(),
-  @Schema(description = "Optional additional information", required = false, defaultValue = "null")
+  @Schema(description = "Optional additional information", nullable = true)
```
xoen, line 15: Wouldn't the default value be `''`without `defaultValue="null"`🤔 ?

This is from the `@Schema` interface:

```
String defaultValue() default "";
```

xoen: Unless `nullable=true` does something clever and sets the default to `null` in this case (which would make sense, but I'm not sure if it does that)

ushkarev: there is no default value as this is a read-only type. default values used only when the field is not provided and therefore something needs to be assumed

xoen (dismissed):

xoen (approved):

PR #173, opened 2024-09-03
IR-431 Add seed configuration and SQL script for local development

Mjwillis:


src/main/resources/db/seed/R__01_docker_seed.sql
```diff
@@ -0,0 +1,31 @@
+DO
```
Mjwillis, line 1: Just some simple sample data for starters

ushkarev (approved):

PR #206, opened 2024-11-12
IR-631: Filter reports by multiple locations

ushkarev:
…and rename `prisonId` fields to `location` in reports and events models to reflect that they do not solely refer to prisons.

Related to [ui#266](https://github.com/ministryofjustice/hmpps-incident-reporting/pull/266)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/Event.kt
```diff
@@ -28,4 +29,10 @@ open class Event(
   val modifiedAt: LocalDateTime,
   @Schema(description = "Username of the person who last changed this event")
   val modifiedBy: String,
-)
+) {
+  // TODO: `prisonId` can be removed once NOMIS reconciliation checks are updated to use `location`
+  @get:Schema(description = "The location where incident took place, typically a NOMIS prison ID", deprecated = true, example = "MDI")
```
Mjwillis, line 34: 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResource.kt
```diff
@@ -84,10 +84,29 @@ class ReportResource(
...
+        arraySchema = Schema(
+          requiredMode = Schema.RequiredMode.NOT_REQUIRED,
+          nullable = true,
+          defaultValue = "null",
+        ),
+      ),
+    )
+    @RequestParam(required = false)
+    location: List<
+      @Size(min = 2, max = 20)
```
Mjwillis, line 101: why must there be at least 2?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResource.kt
```diff
@@ -84,10 +84,29 @@ class ReportResource(
     ],
   )
   fun getBasicReports(
+    @Parameter(
+      description = "Filter by given locations, typically prison IDs",
+      example = "LEI,MDI",
+      array = ArraySchema(
+        schema = Schema(example = "MDI", minLength = 2, maxLength = 20),
```
Mjwillis, line 91: Not sure what the limit should be ! - 20 might be too small - Think we should allow about 150 (e.g. all prisons )

Mjwillis (approved):
Couple of comments. Is there impact on syscon? I think you've kept backward compatibility - but just couldn't be certain!

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResource.kt
ushkarev: this annotation does not affect the size of the list of filters

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResource.kt
ushkarev: this schema is not for the list itself

ushkarev:
>Is there impact on syscon?

well i tried to make this change such that there's none. the old filter remains, the old JSON response field is synthesised by jackson…

PR #208, opened 2024-11-15
IR-577: Removed `QID` prefix from question codes

xoen:
This is not necessary and it may well be in our way when updating answers. Regardless it makes the UI a bit more fiddly because we have to map back and forth between question IDs in the configs and questions code with the prefix (which is not guaranteed to be there for future report types config).

It is now a good time to remove this as data it not impacting real users (only clients are syncing/reconciliation ones)


src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/helper/ReportBuilder.kt
```diff
@@ -103,7 +103,7 @@ fun buildReport(
     )
     (1..generateQuestions).forEach { questionIndex ->
       val historicalQuestion = history.addQuestion(
-        code = "QID-$historyIndex-%012d".format(questionIndex),
+        code = "$historyIndex-%012d".format(questionIndex),
```
ushkarev, line 106: if you're getting rid of the 0 prefix on the question, perhaps this should be simply
```kotlin
"$historyIndex-$questionIndex"
```

xoen: yeah, to be honest I didn't really think it mattered as codes now are "just a string, whatever that is" :)

I guess I can change to remove the padding, but AFAICT this is just a testing machinery rather than something that affect real historic questions' codes right?

ushkarev: it is just a test but… i dunno, it looks odd now that there are 0s in the middle but not at the front

xoen: Updated: a54b269a2fdad8ad34107edc9ff04ec3bbd655e6

ushkarev (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/helper/ReportBuilder.kt
xoen: > it is just a test but… i dunno, it looks odd now that there are 0s in the middle but not at the front

It's just a string, it could be anything, like `waffles-waffles-1-00000-23` 😝  (anyway, I changed, so it shouldn't disturb you any longer :D)

PR #218, opened 2024-11-25
IR-621: Change question-adding endpoint to accept an array of question-with-response objects

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportQuestionResponseResource.kt
```diff
@@ -77,7 +80,7 @@ class ReportQuestionResponseResource(
   @PreAuthorize("hasRole('ROLE_MAINTAIN_INCIDENT_REPORTS') and hasAuthority('SCOPE_write')")
   @ResponseStatus(HttpStatus.CREATED)
```
ushkarev, line 81: currently still a `POST` + 201 `CREATED` request/response cycle but this will change once question code is used for unique constraints

xoen (approved):

PR #221, opened 2024-11-26
IR-696: Add database indices to speed up referencing questions by code

ushkarev:
Improves performance of changes in #220

NB: indices already created in all databases because it takes 120s+ and must be run manually. hence `if not exists` in the migration script.


Mjwillis (approved):

Mjwillis:
Lgtm

PR #222, opened 2024-11-28
IR-698 Prevent transaction contention

Mjwillis:


xoen (approved):

xoen:
I see what this is trying to do and it makes sense...I'm not sure why but a bunch of tests are failing with errors related to transactions now 🤔

PR #225, opened 2024-12-02
IR-699 Step 1: Used sets for collections - to allow equality and full entity graph load

Mjwillis:


build.gradle.kts
```diff
@@ -76,3 +76,7 @@ tasks {
     compilerOptions.jvmTarget = JvmTarget.JVM_21
   }
 }
+
+allOpen {
```
Mjwillis, line 80: Need to support lazy load for sets

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -44,11 +46,28 @@ import java.util.UUID
       name = "Report.eager",
       attributeNodes = [
         NamedAttributeNode("event"),
-        NamedAttributeNode("questions", subgraph = "Report.eager.subgraph"),
+        NamedAttributeNode("staffInvolved"),
```
Mjwillis, line 49: can now load the whole report in 1 sql

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/specifications/Specifications.kt
```diff
@@ -25,20 +26,6 @@ fun <T, V : Comparable<V>> KProperty1<T, V>.buildSpecForLessThan(value: V): Spec
   }
 }
 
-/** Build «less than or equal to» specification from an entity’s property */
```
Mjwillis, line 28: Not used

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/ReportService.kt
```diff
@@ -308,18 +308,22 @@ class ReportService(
             addQuestion(
               code = request.code,
               question = request.question,
+              sequence = this.getQuestions().size,
```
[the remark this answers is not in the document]
ushkarev: other than this, i think everything else in your pr makes sense to me

ushkarev:
also, i'm wondering if there could be a nice way to reduce boilerplate and chances of errors by somehow teaching the entities to expose the important fields as a list or whatever 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Response.kt
```diff
@@ -19,7 +22,7 @@ class Response(
   @ManyToOne(fetch = FetchType.LAZY)
   val question: Question,
 
-  // TODO: should we add a `val code: String` like in Question?
```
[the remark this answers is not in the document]
Mjwillis: Yes we could add the NOMIS response ID - I agree

Mjwillis:
I'm going to merge this and work on the changes you've asked for.  I've found the problem with the constraint.  It's based on the way NOMIS does updates. It's to do with the fact that there is a transitory constraint validation which resolved by commit time but postres is applying the contraint mid-transaction.

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/PrisonerInvolvement.kt
```diff
@@ -52,4 +63,34 @@ class PrisonerInvolvement(
...
+    if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false
+
+    other as PrisonerInvolvement
+
+    if (report != other.report) return false
+    if (id != other.id) return false
+    if (prisonerNumber != other.prisonerNumber) return false
+    if (prisonerRole != other.prisonerRole) return false
+    if (outcome != other.outcome) return false
+    if (comment != other.comment) return false
```
xoen, line 78: I'm looking at your video and when you change the comment the record seemly is updated correctly.

My understanding is that things like `prisonerRole`, `outcome` and comment are now used to determine if it's the same record (`equals()`).

But does that mean that if prisoner `A11111AA` is involved as `PERPETRATOR` but then the value in the upsert request changes it to be `SUSPECTED_INVOLVED` they would be considered as two different entities and therefore a separate record would be created for `A11111AA`? (instead of updating the existing record)

It seems to work in your video so I am missing something here (but tbf I haven't finished watching it yet).

xoen: I think I may start to get it.

Because we are not provided with the primary key in the upsert request, my understanding is that we'll delete the existing record if different and insert a new one.

So effectively anything which doesn't match would trigger an "update" (not in the DB sense but still the result is the same, old state is deleted, new state is saved)

PR #226, opened 2024-12-02
IR-699 Step 2 : Prevent additional deletes when updating questions

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -297,63 +299,89 @@ class Report(
...
-    questions.clear()
-    addNomisQuestions(upsert.questions)
+    updateQuestionAndResponses(upsert.questions)
 
-    history.clear()
+    this.history.clear()
     addNomisHistory(upsert.history)
   }
 
+  private fun updateQuestionAndResponses(nomisQuestions: List<NomisQuestion>) {
```
Mjwillis, line 358: This is the bit that either updates or creates and removed any left over

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Response.kt
```diff
@@ -51,20 +51,23 @@ class Response(
 
     if (question != other.question) return false
     if (sequence != other.sequence) return false
+    if (response != other.response) return false
```
Mjwillis, line 54: Added based on last review

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/specifications/Specifications.kt
```diff
@@ -34,7 +34,7 @@ fun <T, V : Comparable<V>> KProperty1<T, V>.buildSpecForGreaterThanOrEqualTo(val
 }
 
 /** Build «equal to» specification joining to a related entity (via a list property) */
-fun <T, R, V> KProperty1<T, SortedSet<R>>.buildSpecForRelatedEntityPropertyEqualTo(property: KProperty1<R, V>, value: V): Specification<T> {
+fun <T, R, V> KProperty1<T, Collection<R>>.buildSpecForRelatedEntityPropertyEqualTo(property: KProperty1<R, V>, value: V): Specification<T> {
```
Mjwillis, line 37: added based on last review

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -297,63 +299,89 @@ class Report(
...
+    this.correctionRequests.clear()
     addNomisCorrectionRequests(upsert.requirements)
 
-    questions.clear()
-    addNomisQuestions(upsert.questions)
+    updateQuestionAndResponses(upsert.questions)
 
-    history.clear()
+    this.history.clear()
     addNomisHistory(upsert.history)
```
Mjwillis, line 355: Should look at doing the same for history next

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/NomisSyncResourceTest.kt
```diff
@@ -1568,6 +1569,168 @@ class NomisSyncResourceTest : SqsIntegrationTestBase() {
           WhatChanged.ANYTHING,
         )
       }
+
+      @Test
```
Mjwillis, line 1573: This test failed with the constraint on. Although I've remove the constraint now as we have some bad data need fixing first.

src/main/resources/db/migration/V1_6__remove_question_constraints.sql
```diff
@@ -0,0 +1,4 @@
+
+ ALTER TABLE question DROP CONSTRAINT question_code_uniq;
```
Mjwillis, line 2: I've removed the constraints.

ushkarev, line 2: since the constraint was already removed in `prod`, this won't deploy. i would suggest dropping it manually in `dev` for testing and putting it back in manually.

constraints have to be put in manually anyway (since k8s kills pods before they can be applied)

Mjwillis: Should have had DEFERRED on the end of the original

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
ushkarev: there were more than a dozen duplicate historical questions

src/main/resources/db/migration/V1_6__remove_question_constraints.sql
Mjwillis: Yes that's fine

Mjwillis: @ushkarev I've remove the file to remove!

src/test/resources/related-objects/staff-involved/add-response-many.json
```diff
@@ -1,4 +1,9 @@
 [
+  {
```
Mjwillis, line 2: as ID is now removed from equality

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/History.kt
```diff
@@ -39,6 +41,43 @@ class History(
   val questions: SortedSet<HistoricalQuestion> = sortedSetOf(),
 ) : Comparable<History> {
 
+  fun findQuestion(code: String, sequence: Int) = this.questions.firstOrNull { it.code == code && it.sequence == sequence }
```
xoen, line 44: Do we need to consider the sequence? Isn't a question "equal" to another if their codes match?

Sorry, maybe I'm missing something fundamental here (still looking at this PR and the prior one)

Mjwillis: In NOMIS

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -297,63 +324,89 @@ class Report(
...
+        question.updateResponses(nomisQuestion.answers)
+        question
+      }.toSet(),
+    )
+  }
 
-    history.clear()
-    addNomisHistory(upsert.history)
+  fun updateHistory(nomisHistory: Collection<NomisHistory>) {
+    nomisHistory.forEach { history ->
```
ushkarev, line 386: i'm a little confused because different sorted sets of entities are handled differently in the methods here.

should this not have a `this.history.retainAll(…)`? i get that it's a sorted set so if date and type match, a new history item _ought_ not be created, but what about the historical question/response children?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
Mjwillis: I've done all the historical ones now

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
Mjwillis: This was fixed

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/Mapping.kt
```diff
@@ -43,93 +46,71 @@ fun NomisReport.toNewEntity(): Report {
...
-        )
-      }
-  }
+fun Report.updateNomisPrisonerInvolvements(offenderParties: Collection<NomisOffenderParty>) {
+  val newInvolvements = offenderParties.map { offenderParty ->
+    val newPrisoner = createPrisonerInvolvement(offenderParty)
+    prisonersInvolved.find { it == newPrisoner } ?: addPrisonerInvolved(newPrisoner)
+  }.toSet()
+
+  this.prisonersInvolved.retainAll(newInvolvements)
```
ushkarev, line 90: are the retain-all calls a bit of a jpa fudge? the javadocs suggest that only elements *already* in the mutable set are kept. does it mean it's bit of an accident that any newly-created entity actually get saved?

or is the jpa/hibernate design such that instantiating an entity is enough for it to be inserted into db without needing to actually be in this sorted set?

Mjwillis: no it needs to be in the sorted set.  None of this is JPA, it just set theory

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Response.kt
Mjwillis: I've done as you have asked!

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/Mapping.kt
```diff
@@ -85,17 +85,17 @@ fun Report.addNomisCorrectionRequests(correctionRequests: Collection<NomisRequir
 }
 
 fun Report.addNomisQuestions(questions: Collection<NomisQuestion>) {
-  questions.sortedBy { it.sequence }.forEach { question ->
+  questions.forEach { question ->
     val dataItem = this.addQuestion(
       code = question.questionId.toString(),
+      sequence = question.sequence - 1,
```
xoen: Why is the sequence reduced by one?

Mjwillis: NOMIS starts from 1

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/Mapping.kt
xoen: I struggled to understand this at first but then I realised the `addPrisonerInvolved(newPrisoner)` bit was adding to `this.prisonersInvolved` rather than creating a DB record (it makes sense, but I didn't quite get that at first)

ushkarev: right, but there was a point when the entities were [not being added](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/226/commits/e6467fe4709c978d881dc32d4117eac8403761dc)

my question was _only_ about the use of `SortedSet.retainAll`

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/Mapping.kt
xoen: so this is simply to try and have sequences starting from 0 in our DB 🤔 - but nothing to do with the data problems 👍

ushkarev: well we gotta put _something_ there now that ordering must be manually managed

PR #233, opened 2024-12-10
IR-720: OpenAPI annotation fixes

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/NomisSyncResource.kt
```diff
@@ -72,10 +73,16 @@ class NomisSyncResource(
     ],
   )
   fun upsertIncidentReport(
-    @Schema(
-      description = "Incident report created/updated in NOMIS",
-      accessMode = Schema.AccessMode.WRITE_ONLY,
-      oneOf = [NomisSyncRequest::class, NomisSyncCreateRequest::class, NomisSyncUpdateRequest::class],
```
ushkarev, line 78: it seems that generated code for `NomisSyncRequest` from the OpenAPI spec was no longer marking the id field as nullable and optional

xoen (approved):

PR #234, opened 2024-12-11
IR-716: Expose entity fields publicly

ushkarev:
…since there wasn't much reason to hide a small subset of them. Several were exposed via custom getter methods anyway.

For instance both

```kotlin
class Entity(
  private val report: Report
) {
  fun getReport() = report
}
```

and

```kotlin
class Entity(
  val report: Report
)
```

result in the same public getter method `fun getReport(): Report` being synthesised.


xoen (approved):

src/test/resources/entity-mapping/sample-report-with-details.json
```diff
@@ -273,24 +273,28 @@
   ],
   "staffInvolved": [
     {
+      "sequence": 0,
```
xoen, line 276: shouldn't make a difference (I don't think) but aren't sequences starting from `1` now to match NOMIS?

ushkarev: apparently not according to #235

PR #237, opened 2024-12-13
IR-722: Store names of prisoners and staff involved in the database

ushkarev:
1) so that clients of the api don’t require an additional lookup for every view
2) because NOMIS already provides them in sync payloads
3) and (if ever needed) it will enable us to do richer searches in future and to make staff username or prisoner number optional (to support situations where incomplete information is known by reporting officers)

First and last names might as well be separate fields as in NOMIS payloads.

At first, these must be null to allow a migration in every environment without junk data. Once all apis are updated and every environment has had a full NOMIS migration, we will need to make them non-null in database and all api classes


Mjwillis (approved):
LGTM - if a prisoner is not found will the merge fail?

Mjwillis (dismissed):
LGTM

Mjwillis (approved):
LGTM

PR #238, opened 2024-12-13
IR-724 Null out incidentChangeDate

Mjwillis:


src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/NomisSyncResourceTest.kt
```diff
@@ -284,7 +284,6 @@ class NomisSyncResourceTest : SqsIntegrationTestBase() {
             description = "Damage",
             createDateTime = now,
             createdBy = reportingStaff.username,
-            incidentChangeDate = now.toLocalDate(),
```
Mjwillis, line 287: We can remove as we now will fall back to createDateTime if now populated

karenmillermoj (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/NomisHistory.kt
```diff
@@ -17,7 +17,7 @@ data class NomisHistory(
   val questions: List<NomisHistoryQuestion>,
   @Schema(description = "When the questionnaire was changed")
   @Deprecated("Use incidentChangeDateTime instead")
-  val incidentChangeDate: LocalDate,
+  val incidentChangeDate: LocalDate? = null,
```
xoen, line 20: Is the sync service always passing the `incidentChangeDateTime` field now? I do wonder if it's worth removing this deprecated field altogether instead of just making it nullable?

But I may miss something, sorry.

karenmillermoj: ah true - we need to sync pushing to prod, but yes, you're right

PR #241, opened 2024-12-17
IR-697: Permit deleting questions and responses from reports by code

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -423,16 +424,20 @@ class Report(
...
+  fun removeQuestion(question: Question) {
+    questions.remove(question)
+  }
+
+  fun removeQuestions(questionCodes: Set<String>) {
+    val questionsToRemove = questions.filter { questionCodes.contains(it.code) }
+    val questionCodesFound = questionsToRemove.mapTo(mutableSetOf()) { it.code }
+    val missingCodes = questionCodes - questionCodesFound
+    if (missingCodes.isNotEmpty()) {
+      throw QuestionsNotFoundException(missingCodes)
```
xoen, line 436: What does this mean? Does it mean that if the client for whatever reason pass a question ID which is not answered it would get an error?

Is this necessary? Couldn't that just be ignored? I don't think there is any harm. I don't think I would pass anything extra but it shouldn't make any difference (e.g. it's a no-op to remove a question which was not answered before)

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportQuestionResponseResource.kt
```diff
@@ -174,12 +175,30 @@ class ReportQuestionResponseResource(
       ),
     ],
   )
-  fun deleteLastQuestionAndResponses(
+  fun deleteQuestionsAndResponses(
     @Schema(description = "The incident report id", example = "11111111-2222-3333-4444-555555555555", requiredMode = Schema.RequiredMode.REQUIRED)
     @PathVariable
     reportId: UUID,
+    @Parameter(
```
xoen, line 182: So this is not in the request body? 🤔

ushkarev: yeah query string as we discussed a couple of weeks ago

xoen (approved):

PR #250, opened 2024-12-19
IR-735 Remove sending events for NOMIS sync

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/NomisSyncService.kt
```diff
@@ -53,9 +54,9 @@ class NomisSyncService(
   }
 
   private fun updateExistingReport(reportId: UUID, incidentReport: NomisReport): ReportWithDetails {
-    log.debug("Locking existing report: ${incidentReport.incidentId}")
+    log.info("Locking existing report: ${incidentReport.incidentId}")
```
xoen: Should we log the reportID/UUID instead? I feel like it may be easier to investigate via the UI.

ushkarev (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/NomisSyncService.kt
Mjwillis: Yes probably but the ref Id is also nice

ushkarev: easy enough to look in db

PR #254, opened 2024-12-20
IR-738: Remove reasons for correction requests

ushkarev:
This field is not available in NOMIS data and does not appear necessary given progressing designs

See also: [ui#300](https://github.com/ministryofjustice/hmpps-incident-reporting/pull/300)


src/test/resources/related-objects/correction-requests/update-request-partial.json
```diff
@@ -1,3 +1,3 @@
 {
-  "descriptionOfChange": "Description needs to mention prisoner numbers"
+  "descriptionOfChange": "Description needs to mention prisoner name"
```
ushkarev, line 2: annoyingly, the partial update test case is basically the same as the normal one.
…but it'd take a bigger change to the abstract test classes to make partial update test cases optional for related entities.

xoen (approved):

PR #260, opened 2025-01-10
IR-747: Support filtering by references

ushkarev:


jonesj2193 (approved):
From what I can tell it looks good and makes sense to me.

PR #262, opened 2025-01-17
IR-761: Flags to accommodate “skip for now” functionality in UI

ushkarev:
Add flags to report (returned only when requesting full details) that are used to indicate that adding staff or prisoner involvement is incomplete. A freshly created report has both set to true and adding involvements automatically flips the flag to false. However removing all existing involvements will _not_ set the flag back to incomplete.

This will be used only by the UI (to allow users to remind themselves that involvements need to be added later) and downstream data consumers should ignore them.


Mjwillis (approved):
LGTM

PR #267, opened 2025-01-31
IR-771: Copy constant enumerations in to database tables

ushkarev:
…for use only in reporting; Analytical Platform & DPR.

The application will continue to only use all values solely from internal constants in `uk.gov.justice.digital.hmpps.incidentreporting.constants`.

NB:
- any changes to constants / enumeration classes _require_ new migrations!
- these tables should NOT be used in foreign key constraints! otherwise migrations would be overly-complicated
- NOMIS codes are not included because DPR reports are being recreated and there’s no need to preserve a link to data that will be removed


Mjwillis (approved):
LGTM!

PR #271, opened 2025-02-12
IR-778 Wire in the DPR library and add sample reports and tests

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/config/ResourceServerConfiguration.kt
```diff
@@ -2,11 +2,47 @@ package uk.gov.justice.digital.hmpps.incidentreporting.config
...
+      customizer.authorizeHttpRequestsCustomizer.dsl
+        // override the entire authorizeHttpRequests DSL
+        ?.also { dsl -> dsl.invoke(this) }
+        // apply specific customizations to the default authorizeHttpRequests DSL
+        ?: also {
+          customizer.unauthorizedRequestPathsCustomizer.unauthorizedRequestPaths.forEach { authorize(it, permitAll) }
+          customizer.anyRequestRoleCustomizer.defaultRole
+            ?.also { authorize(anyRequest, hasRole(it)) }
+            ?: also {
+              authorize("/reports/**", hasRole("VIEW_INCIDENT_REPORTS"))
```
Mjwillis, line 34: Hack to make the reports secure with a role

src/main/resources/reports/incident-report.json
```diff
@@ -0,0 +1,291 @@
...
+      "connection": "postgres"
+    }
+  ],
+  "policy": [
+    {
+      "id": "role-based-policy",
+      "type": "access",
+      "rule": [
+        {
+          "effect": "permit",
```
Mjwillis, line 23: Had to remove the policy's as we need to use the DPR token to make this work

Mjwillis: Should be 
```json
"policy": [
    {
      "id": "access",
      "type": "access",
      "rule": [
        {
          "effect": "permit",
          "condition": [
            {
              "match": ["${role}", "ROLE_VIEW_INCIDENT_REPORTS"]
            }
          ]
        }
      ]
    },
    {
      "id": "caseloads",
      "type": "row-level",
      "action": ["location IN (${caseloads})"],
      "rule": [
        {
          "effect": "permit",
          "condition": [
            {
              "exists": ["${caseloads}"]
            }
          ]
        }
      ]
    }
  ],```

File: src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/config/ResourceServerConfiguration.kt (hunk 1 above)
Comment (ushkarev) [reply src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/config/ResourceServerConfiguration.kt:34]: is it temporary? might be worth leaving a to-do message in the code because i wouldn't know which part of this bean method would need to remain and what will need to be removed…?

Comment (ushkarev) [reply src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/config/ResourceServerConfiguration.kt:34]: how does this interact with the role annotations elsewhere, do these `authorize(…)` statements replace them?

File: src/main/resources/application-dev.yml [hunk 3]
```diff
@@ -5,7 +5,8 @@ api:
   base:
     url:
       oauth: http://localhost:9090/auth
-      prisoner-search: http://localhost:8081
```
ushkarev, line 8: do you need to change this constant to match?

https://github.com/ministryofjustice/hmpps-incident-reporting-api/blob/8cad7915df43e3b2b70312c57d117a15a73b8a6b/src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/integration/wiremock/PrisonerSearchMockServer.kt#L9

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/integration/wiremock/HmppsAuthMockServer.kt
```diff
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.incidentreporting.integration.wiremock
 
 import com.github.tomakehurst.wiremock.client.WireMock.aResponse
+import com.github.tomakehurst.wiremock.client.WireMock.get
```
ushkarev, line 4: what does this do? doesn't ktlint complain about an unused import?

ushkarev (commented):
can't really say much about the json definitions given i've never seen them in action nor read the documentation 😉

Mjwillis:
@ushkarev would you approve anyway?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/config/ResourceServerConfiguration.kt
Mjwillis: No they supliment them - its just that DPR don't protect their endpoints will a role

Mjwillis: I'll add a comment

src/main/resources/application-dev.yml
Mjwillis: they match already

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/integration/wiremock/HmppsAuthMockServer.kt
Mjwillis: fixed

src/main/resources/application-dev.yml
Mjwillis: I've made everything match both docker, dev and wiremock

ushkarev (approved):

PR #276, opened 2025-02-25
IR-778 Update role permissions and enhance report policy to restrict by caseload

Mjwillis:


src/main/resources/application.yml
```diff
@@ -105,4 +105,4 @@ dpr:
       path: /users/me/caseloads
   endpoint:
     api:
-      role: INCIDENT_REPORTS__RO
+      roles: INCIDENT_REPORTS__RO,INCIDENT_REPORTS__RW,INCIDENT_REPORTS__APPROVE,INCIDENT_REPORTS__PECS
```
ushkarev, line 108: if this is a list of roles where _any_ permits access, then i’d not include `INCIDENT_REPORTS__PECS`

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/config/ResourceServerConfiguration.kt
```diff
@@ -32,18 +34,16 @@ class ResourceServerConfiguration(
           customizer.anyRequestRoleCustomizer.defaultRole
             ?.also { authorize(anyRequest, hasRole(it)) }
             ?: also {
-              // TODO: This has to be a user role - not ideal but will secure with this for now
-              authorize("/report/**", hasRole(dprEndpointApiRole))
-              authorize("/reports/**", hasRole(dprEndpointApiRole))
-              authorize("/definitions/**", hasRole(dprEndpointApiRole))
-              authorize("/statements/**", hasRole(dprEndpointApiRole))
+              authorize("/report/**", hasAnyRole(*dprEndpointApiRoles.toTypedArray()))
```
Mjwillis, line 37: Restrict to all the roles that a user can have - I would have preferred to restrict the endpoints using a system token, but that requires DPR library to check use roles in a different way.  For now I've protected the API endpoints by the same roles as the IRS service.

ushkarev (approved):
looks legit i guess ;)
except maybe the PECS role

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/config/ResourceServerConfiguration.kt
```diff
@@ -32,18 +34,16 @@ class ResourceServerConfiguration(
...
+              authorize("/definitions/**", hasAnyRole(*dprEndpointApiRoles.toTypedArray()))
+              authorize("/statements/**", hasAnyRole(*dprEndpointApiRoles.toTypedArray()))
               authorize(anyRequest, authenticated)
             }
         }
     }
     oauth2ResourceServer {
-      // TODO: will allow override of the DprAuthAwareTokenConverter
-      jwt { jwtAuthenticationConverter = AuthAwareTokenConverter() }
+      jwt { jwtAuthenticationConverter = DefaultDprAuthAwareTokenConverter(caseloadProvider) }
```
Mjwillis, line 46: This is the magic line I couldn't add before without the DPR library change yesterday

src/main/resources/application.yml
Mjwillis, line 108: Added all the roles that are allow in service - yuk!

src/main/resources/reports/incident-count.json
```diff
@@ -16,14 +16,38 @@
...
+              ]
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "id": "caseloads",
+      "type": "row-level",
+      "action": ["location IN (${caseloads})"],
```
Mjwillis, line 40: This filters by caseload

src/main/resources/reports/incident-count.json
```diff
@@ -1111,7 +1141,11 @@
             "formula": "format_date(${start_date}, 'dd/MM/yyyy')",
             "sortable": true,
             "visible": "true",
-            "defaultsort": true
+            "defaultsort": true,
```
Mjwillis, line 1144: Date filters now working (sort of!)

src/main/resources/reports/incident-count.json
```diff
@@ -16,14 +16,38 @@
...
+      "id": "access",
+      "type": "access",
+      "rule": [
+        {
+          "effect": "permit",
+          "condition": [
+            {
+              "match": [
+                "${role}",
+                "ROLE_INCIDENT_REPORTS__RO",
```
Mjwillis, line 28: Protecting the data via a role.  Its questionable whether both the API endpoint AND the call need this restriction, maybe its only needed here, its just I hate "open" API endpoints and HMPPS dev community have agreed we shouldn't generally do that for any data other than reference data

src/main/resources/application.yml
Mjwillis: Yes good point!

PR #280, opened 2025-03-04
IR-818 Update role configuration based on DPR/Kotlin library changes and enhance report filters in JSON schemas

Mjwillis:


src/main/resources/application.yml
```diff
@@ -103,6 +103,5 @@ dpr:
     caseloads:
       host: ${api.base.url.manage.users}
       path: /users/me/caseloads
-  endpoint:
-    api:
-      roles: INCIDENT_REPORTS__RO,INCIDENT_REPORTS__RW,INCIDENT_REPORTS__APPROVE,INCIDENT_REPORTS__PECS
+    user:
+      role: INCIDENT_REPORTS__RO
```
Mjwillis, line 107: Framework only supports 1 role.  Will work with DPR to address this as we really want this protected by a "standard" system role

ushkarev: 😞

ushkarev (approved):

PR #281, opened 2025-03-06
MAP-817 Add exception handling for user authorization failures from DPR and update incident report JSON schema

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ApiExceptionHandler.kt
```diff
@@ -73,6 +76,21 @@ class ApiExceptionHandler {
       )
   }
 
+  @ExceptionHandler(UserAuthorisationException::class)
+  @ResponseStatus(FORBIDDEN)
```
Mjwillis, line 80: DPR auth failures were coming out at 500 in our service

src/main/resources/reports/incident-report.json
```diff
@@ -275,7 +278,8 @@
             "display": "Incident description",
             "sortable": false,
             "visible": "true",
-            "defaultsort": false
+            "defaultsort": false,
+            "wordwrap": "break-words"
```
Mjwillis, line 282: this is useful in Dev

PR #282, opened 2025-03-07
SDIT-2639: 🐛 Update swagger to make openapi code generation happy

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/NomisSyncRequest.kt
```diff
@@ -12,13 +12,11 @@ data class NomisSyncRequest(
     requiredMode = Schema.RequiredMode.NOT_REQUIRED,
     example = "123e4567-e89b-12d3-a456-426614174000",
     nullable = true,
-    defaultValue = "null",
   )
   val id: UUID? = null,
   @get:Schema(
     description = "Set to true for initial migration, omit or set to false for updates",
     requiredMode = Schema.RequiredMode.NOT_REQUIRED,
-    allowableValues = ["true"],
```
petergphillips, line 21: this caused
```
    val initialMigration: NomisSyncRequest.InitialMigration? = false
```
to be generated

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/NomisSyncRequest.kt
```diff
@@ -32,39 +30,3 @@ data class NomisSyncRequest(
     }
   }
 }
-
-@Schema(description = "Incident report created in NOMIS", accessMode = Schema.AccessMode.WRITE_ONLY)
-interface NomisSyncCreateRequest {
```
petergphillips, line 37: interface not used anywhere?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/NomisSyncRequest.kt
```diff
@@ -12,13 +12,11 @@ data class NomisSyncRequest(
     requiredMode = Schema.RequiredMode.NOT_REQUIRED,
     example = "123e4567-e89b-12d3-a456-426614174000",
     nullable = true,
-    defaultValue = "null",
```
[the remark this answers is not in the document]
petergphillips: by making this change code generation then creates
```kotlin
    val id: java.util.UUID? = null,
```
anyway

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/NomisSyncRequest.kt
petergphillips: new generation creates
```kotlin
    val initialMigration: kotlin.Boolean? = false
```
still a bit odd to have a `NOT_REQUIRED` field that isn't nullable in your code, but I've left that alone.

petergphillips: didn't really make sense to limit the `allowableValues` like that anyway when you've got the `defaultValue` to be `false` 😁

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/NomisSyncRequest.kt
ushkarev: it was intended for clearer documentation. `NomisSyncRequest` was the form that the NOMIS migration side was using, but `NomisSyncCreateRequest` & `NomisSyncUpdateRequest` more accurately described the interaction between fields

ushkarev: essentially, iff `id` was present, it's an update.
but an update does not make sense with `initialMigration`=true

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/NomisSyncRequest.kt
```diff
@@ -38,7 +36,6 @@ interface NomisSyncCreateRequest {
   @get:Schema(
     description = "Set to true for initial migration",
     requiredMode = Schema.RequiredMode.NOT_REQUIRED,
-    allowableValues = ["true", "false"],
```
petergphillips, line 41: I've done the minimum here to get the generated code compiling - had to remove these as enums were generated with them in

ushkarev (approved):

PR #299, opened 2025-04-01
IR-959: Type families

ushkarev:
Incident types are now grouped into families which share the same description.

Type codes have been renamed to stick to a sensible pattern: `[family-code]_[version]`

```
ABSCONDER → ABSCOND_1
ASSAULT → ASSAULT_5
ATTEMPTED_ESCAPE_FROM_CUSTODY → ATTEMPTED_ESCAPE_FROM_PRISON_1
ATTEMPTED_ESCAPE_FROM_ESCORT → ATTEMPTED_ESCAPE_FROM_ESCORT_1
BOMB_THREAT → BOMB_1
BREACH_OF_SECURITY → BREACH_OF_SECURITY_1
DAMAGE → DAMAGE_1
DEATH_IN_CUSTODY → DEATH_PRISONER_1
DEATH_OTHER → DEATH_OTHER_1
DISORDER → DISORDER_2
DRONE_SIGHTING → DRONE_SIGHTING_3
ESCAPE_FROM_CUSTODY → ESCAPE_FROM_PRISON_1
ESCAPE_FROM_ESCORT → ESCAPE_FROM_ESCORT_1
FINDS → FIND_6
FIRE → FIRE_1
FOOD_REFUSAL → FOOD_REFUSAL_1
FULL_CLOSE_DOWN_SEARCH → CLOSE_DOWN_SEARCH_1
KEY_LOCK_INCIDENT → KEY_OR_LOCK_2
MISCELLANEOUS → MISCELLANEOUS_1
OLD_ASSAULT → ASSAULT_1
OLD_ASSAULT1 → ASSAULT_2
OLD_ASSAULT2 → ASSAULT_3
OLD_ASSAULT3 → ASSAULT_4
OLD_BARRICADE → BARRICADE_1
OLD_CONCERTED_INDISCIPLINE → CONCERTED_INDISCIPLINE_1
OLD_DISORDER → DISORDER_1
OLD_DRONE_SIGHTING → DRONE_SIGHTING_1
OLD_DRONE_SIGHTING1 → DRONE_SIGHTING_2
OLD_DRUGS → DRUGS_1
OLD_FINDS → FIND_1
OLD_FINDS1 → FIND_2
OLD_FINDS2 → FIND_3
OLD_FINDS3 → FIND_4
OLD_FINDS4 → FIND_5
OLD_FIREARM_ETC → FIREARM_1
OLD_HOSTAGE → HOSTAGE_1
OLD_KEY_LOCK_INCIDENT → KEY_OR_LOCK_1
OLD_MOBILES → MOBILE_PHONE_1
OLD_ROOF_CLIMB → INCIDENT_AT_HEIGHT_1
OLD_TEMPORARY_RELEASE_FAILURE → TEMPORARY_RELEASE_FAILURE_1
OLD_TEMPORARY_RELEASE_FAILURE1 → TEMPORARY_RELEASE_FAILURE_2
OLD_TEMPORARY_RELEASE_FAILURE2 → TEMPORARY_RELEASE_FAILURE_3
RADIO_COMPROMISE → RADIO_COMPROMISE_1
RELEASED_IN_ERROR → RELEASE_IN_ERROR_1
SELF_HARM → SELF_HARM_1
TEMPORARY_RELEASE_FAILURE → TEMPORARY_RELEASE_FAILURE_4
TOOL_LOSS → TOOL_LOSS_1
```


Mjwillis (approved):
LGTM

src/main/resources/db/migration/V1_16__type_families.sql
```diff
@@ -46,3 +46,148 @@ values (0, 'ABSCOND', 'Abscond'),
        (28, 'SELF_HARM', 'Self harm'),
        (29, 'TEMPORARY_RELEASE_FAILURE', 'Temporary release failure'),
        (30, 'TOOL_LOSS', 'Tool or implement loss');
+
+alter table constant_type
+  add column family_code varchar(60);
```
xoen: We can make `family_code` `NOT NULL`-able after `UPDATE` statements below.

ushkarev: this is a fair question and i wasn't sure how this would actually work – but it seems to! i think it fills the column with blank strings

xoen: It wasn't a question.

What I meant is that aside from when the column will be added and there will have no way of populating it I don't think the `family_code` should be `NULL`-able (but of course it can't be `NOT NULL`-able until the `UPDATE` statements have populated this column for all the records)

> i wasn't sure how this would actually work – but it seems to!

The reason why it works is because that `ALTER TABLE constant_type ADD COLUMN family_code ...` statement doesn't specify `NOT NULL` so the new column _**is**_ `NULL`-able (again, this makes sense because there is no sensible way to populate this new column for existing records, hence the `UPDATE` statements below) - but after those statements I think it doesn't make sense for this column to be `NULL`-able (unless I'm missing something of course).

ushkarev: oh!! omg how stupid of me lol

xoen: I don't think you're stupid or anything, there is nothing wrong with this migration at all, the newly added column **has to be**  `NULL`-able to begin with (unless you also provide a `DEFAULT` which negates the benefit of `NOT NULL` in this case).

But IMHO once we populated this column (as you do), I think it's sensible for it to be `NOT NULL`-able - hope I'm making sense.

ushkarev: stupid in the sense that i had it completely backwards 🙈

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/Status.kt
```diff
@@ -19,7 +19,7 @@ enum class Status(
   AWAITING_ANALYSIS("Awaiting analysis", "AWAN"),
   IN_ANALYSIS("In analysis", "INAN"),
   INFORMATION_REQUIRED("Information required", "INREQ"),
-  INFORMATION_AMENDED("Information amened", "INAME"),
+  INFORMATION_AMENDED("Information amended", "INAME"),
```
xoen, line 22: Amen 🙏

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/Type.kt
```diff
@@ -17,55 +17,55 @@ enum class Type(
...
+  CONCERTED_INDISCIPLINE_1(TypeFamily.CONCERTED_INDISCIPLINE, "CON_INDISC", active = false),
+  DISORDER_1(TypeFamily.DISORDER, "DISORDER", active = false),
+  DRONE_SIGHTING_1(TypeFamily.DRONE_SIGHTING, "DRONE", active = false),
+  DRONE_SIGHTING_2(TypeFamily.DRONE_SIGHTING, "DRONE1", active = false),
+  DRUGS_1(TypeFamily.DRUGS, "DRUGS", active = false),
+  FIND_1(TypeFamily.FIND, "FINDS", active = false),
+  FIND_2(TypeFamily.FIND, "FIND", active = false),
+  FIND_3(TypeFamily.FIND, "FIND1", active = false),
+  FIND_4(TypeFamily.FIND, "FIND0322", active = false),
+  FIND_5(TypeFamily.FIND, "FINDS1", active = false),
```
xoen, line 48: This is so much nicer 👍

xoen:
On a separate note, I imagine deploying this would break filtering on the UI (not a huge deal) until that related PR is deployed too right?

ushkarev:
yes, filtering in the dashboard will be a bit broken. it'll list all types so there will be like 6 “Assault” in the list and you cannot tell them apart 🤣
i haven't been able to actually test that page in the ui cos it requires the api to be updated first

xoen:
> yes, filtering in the dashboard will be a bit broken.

Yeah, as I said, not a biggie 👍 

> it'll list all types so there will be like 6 “Assault” in the list and you cannot tell them apart 🤣

Hence my objection for doing the grouping just yet. Maybe a simple enough workaround is to have the code in parenthesis for the time being (e.g. `Find of illicit items (FIND_6)`.

PR #306, opened 2025-04-09
IR-981 Add self harm summary materialized view and update report configuration

Mjwillis:


ushkarev (dismissed):
looks legit :)

but i don’t think baking in question-based logic into SQL here will be sustainable in the long term. for one thing, self harm is a candidate type for a wildly reimagined version 2 post MVP

ushkarev (approved):

Mjwillis:
> looks legit :)
> 
> but i don’t think baking in question-based logic into SQL here will be sustainable in the long term. for one thing, self harm is a candidate type for a wildly reimagined version 2 post MVP

Yes agree @ushkarev  - the views actually make it easier to adapt when business changes occur without having to change the reports.  There are more views I think I'll need to do for finds, assaults and yes I think we made need different versions of these as we bring in new ones

PR #308, opened 2025-04-11
IR-993 Add serious sexual assault summary materialized view and update report configurations

Mjwillis:


src/main/resources/db/views/R__101_serious_sexual_summary.sql
```diff
@@ -0,0 +1,70 @@
+drop materialized view if exists serious_sexual_assault_summary;
+
+create materialized view serious_sexual_assault_summary as
```
ushkarev, line 3: it looks like there will be quite a few of these: i think it'd be nice to prefix the views with some kind of DPR-related prefix so we can tell what they're for. what do you think?

Mjwillis: I'll do that in the next pr

PR #311, opened 2025-04-15
IR-983 Rename self harm and serious sexual assault materialized views and update related queries

Mjwillis:


xoen (approved):
Part of me would have preferred to have `view_` as a prefix even tho, admittedly doesn't sound as good, but maybe this is good enough to distinguish these DB tables from the "real" ones.

jonesj2193 (approved):

PR #312, opened 2025-04-16
IR-1006 Add location field to CorrectionRequest and update related SQL schema

Mjwillis:


src/main/resources/db/migration/V1_18__staff_location_correction.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE correction_request add column location varchar(20);
```
Mjwillis, line 1: This will need to be made not nullable once migrated again

ushkarev: while we still need to figure out the business needs here *first*, i think it ought to remain nullable

PR #313, opened 2025-04-17
IR-887: Description addendums (data model)

xoen:
- a report can have many description addendums (text appended to the description by a user)
- added `description_addendum` table to contain these records
- endpoints returning a `ReportWithDetails` response will include these description addendums (while basic report will not as they don't include these kind of related objects, for performance reasons)
- this should not be a breaking change

NOTE: This is the foundation for other parts of this work, e.g. an endpoint to create these description addendums, but these will be part of separate PRs.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/DescriptionAddendum.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/DescriptionAddendum.kt
@@ -1,0 +1,16 @@
+package uk.gov.justice.digital.hmpps.incidentreporting.dto
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDateTime
+
+@Schema(description = "Addendum to the description", accessMode = Schema.AccessMode.READ_ONLY)
+@JsonInclude(JsonInclude.Include.ALWAYS)
+data class DescriptionAddendum(
+  @Schema(description = "Username of user who added this addendum", example = "USER_1")
+  val createdBy: String,
+  @Schema(description = "When addendum was added", example = "2024-04-29T12:34:56.789012")
+  val createdAt: LocalDateTime,
+  @Schema(description = "Addendum text")
+  val text: String,
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/DescriptionAddendum.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/DescriptionAddendum.kt
@@ -11,6 +11,10 @@ data class DescriptionAddendum(
   val createdBy: String,
   @Schema(description = "When addendum was added", example = "2024-04-29T12:34:56.789012")
   val createdAt: LocalDateTime,
+  @Schema(description = "First name of person that added this addendum", example = "John")
+  val firstName: String,
+  @Schema(description = "Last name of person that added this addendum", example = "Doe")
+  val lastName: String,
   @Schema(description = "Addendum text")
   val text: String,
 )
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/DescriptionAddendum.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDateTime
+
+@Schema(description = "Addendum to the description", accessMode = Schema.AccessMode.READ_ONLY)
+@JsonInclude(JsonInclude.Include.ALWAYS)
+data class DescriptionAddendum(
+  @Schema(description = "Username of user who added this addendum", example = "USER_1")
+  val createdBy: String,
```
Mjwillis, line 11: @xoen we will also want to have the first and last name of the staff member, as the appended data stores their name and NOT their username - to populate this field we are going to have to do a reverse lookup and potentially we either may not find the record or there will be more than one result.  My feeling is that for migrated records we just store the first  and last name and leave the createdBy blank but for new records we can populate username and their name (like we do for the staff involvement )

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/DescriptionAddendum.kt
```diff
@@ -0,0 +1,69 @@
...
+@EntityOpen
+class DescriptionAddendum(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long? = null,
+
+  @ManyToOne(fetch = FetchType.LAZY)
+  val report: Report,
+
+  val createdBy: String,
```
Mjwillis, line 24: Add firstName and lastName

src/main/resources/db/migration/V1_19__add_description_addendum.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE description_addendum
+(
+  id                      SERIAL
+    CONSTRAINT description_addendum_pk PRIMARY KEY,
+  report_id               UUID         NOT NULL
+    CONSTRAINT description_addendum_report_fk REFERENCES report (id) ON DELETE CASCADE,
+  "text"     TEXT         NOT NULL,
+  created_at TIMESTAMP    NOT NULL,
```
Mjwillis, line 8: Need 2 more columns for first and last name

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
```diff
@@ -573,6 +574,18 @@ class ReportResourceTest : SqsIntegrationTestBase() {
               "prisonId": "MDI",
               "title": "Incident Report 11124143",
               "description": "A new incident created in the new service of type find of illicit items",
+              "descriptionAddendums": [
+                {
+                  "createdBy": "staff-1",
```
Mjwillis, line 579: Its worth having their name to make display quicker too

Mjwillis (changes requested):
- add first and last name to entity, and database

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/DescriptionAddendum.kt
xoen: done here: https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/313/commits/3047e4c3a02d81617abe4eda44953f78c04c687a

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
xoen: done here: https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/313/commits/3047e4c3a02d81617abe4eda44953f78c04c687a

Mjwillis (approved):
LGTM

PR #318, opened 2025-04-30
IR-1008 Update refresh-views-cronjob.yaml to adjust non-production schedule to 19:00 and exclude pre-prod

Mjwillis:
This Pull Request (PR) updates the configuration for the refresh-views-cronjob in the hmpps-incident-reporting-api Helm deployment template. The changes introduce environment-specific conditions and modify the cronjob schedule for non-production environments.
### Main Changes
- Added conditional logic to include the cronjob definition only for environments prod and dev.
- Adjusted the schedule for non-production environments to run at 19:00 instead of 20:00.
- Ensured the cronjob definition is enclosed within the new conditional block to avoid unnecessary deployment in other environments.

### DIFF
--- a/helm_deploy/hmpps-incident-reporting-api/templates/refresh-views-cronjob.yaml
+++ b/helm_deploy/hmpps-incident-reporting-api/templates/refresh-views-cronjob.yaml
@@ -1,3 +1,4 @@
+{{- if or (eq .Values.environment.name "prod") (eq .Values.environment.name "dev") }}
 apiVersion: batch/v1
 kind: CronJob
 metadata:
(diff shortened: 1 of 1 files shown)


ushkarev (approved):
why just these 2 envs?

Mjwillis:
> why just these 2 envs?

@ushkarev because pre-prod will get refreshed from prod every 2 weeks anyway - doesn't seem worth it and also it seems to be causing problems with the refresh - I'll disable it for now until I'm sure its not impacting on the refresh and then look at whether its worth re-enabling

PR #323, opened 2025-05-02
IR-1000 Add sequence tracking to description addendums

Mjwillis:
#### Summary
This Pull Request introduces a new sequence field to the DescriptionAddendum entity, which is used to order addendums in incident reports. The changes also include updates to handle this new field in database schemas, DTOs, services, resources, and tests.
### Main Changes
- Added a new sequence field to the DescriptionAddendum entity, DTOs, and database schema for ordering purposes.
- Updated methods across services and repositories to populate and use the sequence field.
- Introduced a new DescriptionAddendumRequest DTO to handle addendum creation through API endpoints.
- Updated sorting logic in the DescriptionAddendum entity to use the sequence field.
- Modified existing tests and added cases to validate the use of the sequence field.
- Updated database migrations to reflect the addition of the sequence column with necessary constraints and indices.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/DescriptionAddendumRequest.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/DescriptionAddendumRequest.kt
@@ -1,14 +1,15 @@
-package uk.gov.justice.digital.hmpps.incidentreporting.dto
+package uk.gov.justice.digital.hmpps.incidentreporting.dto.request
 
 import com.fasterxml.jackson.annotation.JsonInclude
 import io.swagger.v3.oas.annotations.media.Schema
 import java.time.LocalDateTime
 
-@Schema(description = "Addendum to the description", accessMode = Schema.AccessMode.READ_ONLY)
+@Schema(
+  description = "Request to add an addendum to the description of the incident report",
+  accessMode = Schema.AccessMode.READ_ONLY,
+)
 @JsonInclude(JsonInclude.Include.ALWAYS)
-data class DescriptionAddendum(
-  @Schema(description = "Sequence of the addendums for this report", example = "0")
-  val sequence: Int,
+class DescriptionAddendumRequest(
   @Schema(description = "Username of user who added this addendum", example = "USER_1")
   val createdBy: String,
   @Schema(description = "When addendum was added", example = "2024-04-29T12:34:56.789012")
@@ -17,6 +18,6 @@ data class DescriptionAddendum(
   val firstName: String,
   @Schema(description = "Last name of person that added this addendum", example = "Doe")
   val lastName: String,
-  @Schema(description = "Addendum text")
+  @Schema(description = "Addendum text", example = "Correction made to the report")
   val text: String,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/DescriptionAddendum.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/DescriptionAddendum.kt
@@ -21,6 +21,8 @@ class DescriptionAddendum(
   @ManyToOne(fetch = FetchType.LAZY)
   val report: Report,
 
+  val sequence: Int,
+
   var createdBy: String,
   var firstName: String,
   var lastName: String,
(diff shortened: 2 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/DescriptionAddendumRequest.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.v3.oas.annotations.media.Schema
+import java.time.LocalDateTime
+
+@Schema(
+  description = "Request to add an addendum to the description of the incident report",
+  accessMode = Schema.AccessMode.READ_ONLY,
+)
+@JsonInclude(JsonInclude.Include.ALWAYS)
+class DescriptionAddendumRequest(
```
Mjwillis, line 12: I prefer an explicit request object for API

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/DescriptionAddendum.kt
```diff
@@ -31,11 +33,7 @@ class DescriptionAddendum(
   companion object {
     private val COMPARATOR = compareBy<DescriptionAddendum>
       { it.report }
-      .thenBy { it.createdAt }
-      .thenBy { it.createdBy }
-      .thenBy { it.firstName }
-      .thenBy { it.lastName }
-      .thenBy { it.text }
+      .thenBy { it.sequence }
```
Mjwillis, line 36: Using existing pattern for this

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -232,6 +232,7 @@ class Report(
       lastName = lastName,
       createdAt = createdAt,
       text = text,
+      sequence = this.descriptionAddendums.size,
```
Mjwillis, line 235: This is the bit I'm not sure about

src/main/resources/db/migration/V1_20__add_sequence_to_description_addendum.sql
```diff
@@ -0,0 +1,18 @@
+DROP TABLE if exists description_addendum;
```
Mjwillis, line 1: We will lose data temporarily

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
ushkarev: in other "related objects" the entity's add method requires a sequence to be provided and an example of that is

```kotlin
val sequence = if (report.correctionRequests.isEmpty()) 0 else report.correctionRequests.last().sequence + 1
```
…the logic being that should a thing get deleted the size would be smaller than the last used sequence.

but since there's no deletion of description addenda, your route looks fine to me (for now at least!)

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/DescriptionAddendumRequest.kt
ushkarev: i agree. it's what we do for the other "related objects"

src/main/resources/db/migration/V1_20__add_sequence_to_description_addendum.sql
```diff
@@ -0,0 +1,18 @@
...
+
+CREATE TABLE description_addendum
+(
+  id         SERIAL
+    CONSTRAINT description_addendum_pk PRIMARY KEY,
+  report_id  UUID         NOT NULL
+    CONSTRAINT description_addendum_report_fk REFERENCES report (id) ON DELETE CASCADE,
+  sequence   INTEGER      NOT NULL,
```
ushkarev, line 9: other tables with `sequence` have default 0 – no biggy, just wondered if we care?

```sql
sequence               integer default 0 not null
```

ushkarev (approved):

src/main/resources/db/migration/V1_20__add_sequence_to_description_addendum.sql
ushkarev: but it's easier to recreate table than manually run sql as part of deployment and we gotta migrate anyway

PR #324, opened 2025-05-02
IR-1000: Simplify description splitting a tiny bit and increase test coverage

ushkarev:


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/nomis/NomisReport.kt
```diff
@@ -11,7 +11,6 @@ import java.time.format.DateTimeFormatterBuilder
 
 private const val DATETIME_PATTERN = "\\d{2}-[A-Z]{3}-\\d{4} \\d{2}:\\d{2}"
 private val DATETIME_PATTERN_REGEX = DATETIME_PATTERN.toRegex()
-private val USER_REGEX = "User:".toRegex()
```
ushkarev, line 14: this can be a plain string

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/GetDescriptionPartsTest.kt
```diff
@@ -99,65 +99,77 @@ class GetDescriptionPartsTest {
...
-
-    assertThatThrownBy { minimalReportDto.getDescriptionParts() }
-      .isInstanceOf(ValidationException::class.java)
-      .hasMessage(
-        "Validation issue: " +
-          "STARK,TONY Date:Some updated details",
-      )
+    @Test
+    fun `can accept null description`() {
+      val nomisReport = createBasicReport(null)
```
ushkarev, line 105: technically, description can be null and it’s handled: so we ought to test for that too

jonesj2193 (approved):

PR #329, opened 2025-05-07
IR-1000: Add full CRUD endpoints for description addenda

ushkarev:
… so that the REST api can be used the same way for all related objects (like prisoner involvements, etc)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportDescriptionAddendumResource.kt
```diff
@@ -0,0 +1,265 @@
...
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "404",
+        description = "Data not found",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  @Transactional(readOnly = true)
```
Mjwillis, line 60: I notice this pattern in the service with having the transaction in the resource, this isn't ideal, and also non-standard from other APIs.  The transaction boundary should be at the service level.  Its not a problem yet as we don't have any listening clients to events but when we do the data will potentially not yet be committed when the event fires.  Also if the transaction fails the event will be emitted and the data never updated

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportDescriptionAddendumResource.kt
```diff
@@ -0,0 +1,265 @@
...
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "404",
+        description = "Data not found",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+    ],
+  )
+  @Transactional
```
Mjwillis, line 174: Yes ditto with this `@Transactional` - they shouldn't been on the resources

Mjwillis (approved):
Ok, but the main problem is the `@Transactional` which should not be on rest endpoint controllers.  All transactions begin in the service layer

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportDescriptionAddendumResource.kt
```diff
@@ -0,0 +1,265 @@
...
+        report.addDescriptionAddendum(
+          sequence = sequence,
+          createdBy = createdBy ?: authenticationHolder.username ?: SYSTEM_USERNAME,
+          createdAt = createdAt ?: LocalDateTime.now(clock),
+          firstName = firstName,
+          lastName = lastName,
+          text = text,
+        )
+      }
+      report.descriptionAddendums.map { it.toDto() }
```
Mjwillis, line 136: you need to emit an event for any change

Mjwillis (commented):
Emit events for data changes in all cases

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportDescriptionAddendumResource.kt
ushkarev: the abstract super class does this. note the `updateReportOrThrowNotFound` on like 121 (and for all other object-modifying methods)

Ref: https://github.com/ministryofjustice/hmpps-incident-reporting-api/blob/70f5d3a230596448710482201788ef55080a5422/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportRelatedObjectsResource.kt#L58-L64

PR #330, opened 2025-05-08
Fix transaction annotations

ushkarev:


Mjwillis (approved):
All looks good, nice you didn't have to change the test

ushkarev:
> nice you didn't have to change the test
i tried to use that as proof that no logic was changed!

PR #332, opened 2025-05-15
IR-1041: Rename statuses

xoen:
Renamed statuses:
- `AWAITING_ANALYSIS` => `AWAITING_REVIEW`
- `IN_ANALYSIS` => `ON_HOLD`
- `INFORMATION_REQUIRED` => `NEEDS_UPDATING`
- `INFORMATION_AMENDED` => `UPDATED`

Added:
- `NOT_REPORTABLE`
- `REOPENED`
- `WAS_CLOSED`


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/Status.kt
```diff
@@ -16,14 +16,22 @@ enum class Status(
...
-  IN_ANALYSIS("In analysis", "INAN"),
-  INFORMATION_REQUIRED("Information required", "INREQ"),
-  INFORMATION_AMENDED("Information amended", "INAME"),
+  AWAITING_REVIEW("Awaiting review", "AWAN"),
+  ON_HOLD("On hold", "INAN"),
+  NEEDS_UPDATING("Needs updating", "INREQ"),
+  UPDATED("Updated", "INAME"),
   CLOSED("Closed", "CLOSE"),
+
+  // TODO: POST_INCIDENT_UPDATE/INCIDENT_UPDATED will be removed
```
ushkarev, line 25: fyi `INCIDENT_UPDATED` is gone from `preprod` and `prod` but not `dev` so annoyingly can’t be removed yet

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/Status.kt
```diff
@@ -16,14 +16,22 @@ enum class Status(
...
   POST_INCIDENT_UPDATE("Post-incident update", "PIU"),
   INCIDENT_UPDATED("Incident updated", "IUP"),
+
   DUPLICATE("Duplicate", "DUP"),
+
+  // TODO: New statuses tbc
```
ushkarev, line 31: i think we should take this moment to reorder statuses into some kind of order. do u think this is as sensible as can be?

xoen: Tbh when I had a think at this some time ago I ended up with the conclusion there isn't such a natural order aside from "DRAFT" is first, "CLOSED" is last 😅 - anything else could really happen in almost any order so I am not quite sure.

I'm tempted to think, it doesn't matter (It shouldn't right?)

> do u think this is as sensible as can be?

I think it kinda does actually, the first block is pretty much the right order, then you got the 2 statuses that will go at some point and then you have the "not reportable" statuses which I think they can go after "CLOSED".

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/Status.kt
Mjwillis: I'm going to remove from dev both post incident and updated statuses in dev now

Mjwillis (approved):
LGTM

ushkarev (dismissed):
bear in mind that migrations cannot be changed later so is this note something you're ok staying permanently?
```sql
-- TODO: What happens to statuses removed in NOMIS?
--       (e.g. 'Post incident update' and 'incident updated')
--       Does sync takes care of it? Do we need another migration
--       to update existing reports/history?
```
also:
1) sync doesn't take care of it – we will keep receiving `PIU` (`POST_INCIDENT_UPDATE`) status from nomis
2) as of this afternoon, we can remove `IUP` (`INCIDENT_UPDATED`) cos it's now gone from dev nomis

xoen:
@ushkarev I've remove the comment in the migration.

ushkarev (approved):

PR #335, opened 2025-05-20
IR-1040: Removed `INCIDENT_UPDATED` status

xoen:
Mike said occurrences have been removed from all
systems, including NOMIS. I checked `preprod` DB
and I didn't see any rows in `report` or `status_history` tables with this status.

### DIFF
--- a/src/main/resources/db/migration/V1_22__delete_incident_updated_status.sql
+++ b/src/main/resources/db/migration/V1_22__delete_incident_updated_status.sql
@@ -1,0 +1,6 @@
+-- Removes INCIDENT_UPDATED status
+-- See: https://dsdmoj.atlassian.net/wiki/spaces/IR/pages/5589893167/Status+old+and+new
+
+DELETE FROM constant_status
+WHERE code = 'INCIDENT_UPDATED';
+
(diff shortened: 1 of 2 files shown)


src/main/resources/db/migration/V1_22__delete_incident_updated_status.sql
```diff
@@ -0,0 +1,6 @@
+-- Removes INCIDENT_UPDATED status
+-- See: https://dsdmoj.atlassian.net/wiki/spaces/IR/pages/5589893167/Status+old+and+new
+
+DELETE FROM constant_status
+WHERE code = 'INCIDENT_UPDATED';
```
ushkarev, line 5: i would have updated the sequence numbers

xoen: The order is retained, no point in messing with the sequences.

Mjwillis (approved):
LGTM - assume there is an front end one for this too

xoen:
> LGTM - assume there is an front end one for this too
I think it's just the update of the `statuses.ts` and tweak of a comment. Should follow shortly.

PR #337, opened 2025-05-23
IR-1068 Add SCHEMA_TO_RESTORE variable for selective database restoration

Mjwillis:
```
### Description

This pull request introduces an environment variable `SCHEMA_TO_RESTORE` to control schema restoration during database recovery. Specifically, the updates enable only the `public` schema to be restored while excluding the `report` schema.


```


ushkarev (approved):
does the restore doodah already support this var?

PR #347, opened 2025-06-18
Migration to GitHub Actions (from CircleCI)

xoen:


Mjwillis (approved):
LGTM

xoen:
@Mjwillis the CircleCI check now fails because I've removed its configuration, once we merge this and I've tested deployment also work I can remove the old checks (and replace them with the GHA ones)

ushkarev (commented):
i still maintain that we're better off using the executor with pg & localstack sidecars

ushkarev (approved):

PR #350, opened 2025-06-30
Adding 'label' to questions and responses

jonesj2193:
This Pull Request introduces a new label field for both questions and responses, to various response-related models, both in the API layer (DTOs) and the database schema. The label field serves to represent what users saw in the DPS service when saving their response, enabling a full recreation of what existed at the time. 

### Main Changes:
New Field Addition:

- Added label as a field for both question and response level in the HistoricalQuestion, Question, HistoricalResponse, Response and related models.
- Updated constructors and methods across models to incorporate the new label fields.

Database Updates:

- Added a new label column to the question, historical_question, response and historical_response tables via SQL migration.

Logic Updates:

- Modified methods to handle the new label fields for inserting/updating questions in entities like Question, HistoricalQuestion, Response and HistoricalResponse.

Test Updates:

- Updated various test files (e.g., requests, responses, and integration tests) to include the code field in the test data for better validation.


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/HistoricalQuestion.kt
```diff
@@ -16,6 +16,8 @@ data class HistoricalQuestion(
   val code: String,
   @Schema(description = "The question text as seen by downstream data consumers")
   val question: String,
+  @Schema(description = "The question text as seen by the users in DPS", nullable = true)
+  val questionLabel: String,
```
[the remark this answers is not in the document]
xoen: The fact this is a label for a question should be implicit by the context, e.g. we the `code` fields is not called `questionCode`/`responseCode`...speaking of responses we should add the responses label too no?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/Question.kt
```diff
@@ -13,6 +13,8 @@ data class Question(
   val code: String,
   @Schema(description = "The question text as seen by downstream data consumers")
   val question: String,
+  @Schema(description = "The question text as seen by the users in DPS", nullable = true)
```
xoen, line 16: The documentation mentions this is the text seen by users **in DPS**. It made me wonder if that's the actual case?

This is mandatory now (which would break UI and also NOMIS sync, is that what we want?) so it means for reports created in NOMIS it would be the version of the question/response in NOMIS (e.g. likely in ALL-CAPS and all of that).

While presumably the UI will send the better/improved label when updating the questions?

Unless I'm missing something of course, if that's the case maybe drop `in DPS` would be more accurate and less confusing?

src/main/resources/db/migration/V1_25_add_question_label_to_questions.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE question ADD COLUMN question_label VARCHAR(60);
```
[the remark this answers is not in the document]
jonesj2193: Ah this one, I copied it across from another PR and changed the variables and tables, and didn't think of/missed the character limit number.

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Report.kt
```diff
@@ -597,6 +600,7 @@ class Report(
       val historicalQuestion = history.addQuestion(
         code = question.code,
         question = question.question,
+        questionLabel = question.questionLabel.toString(),
```
xoen, line 603: Why `toString()` isn't it a string already? 🤔

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/Question.kt
jonesj2193: So I may be wrong on this, but I think they are intended to be nullable as the NOMIS side won't populate these fields? and then once we have migrated over, that's why there is a TODO, we will make it not nullable and mandatory. But I may be mistaken/misunderstood something along the way

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/HistoricalQuestion.kt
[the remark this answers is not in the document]
xoen: I would 100% definitely call this just `label` tho, there is no reason for this to be more explicit as the context (e.g. you're in a question or response instance) makes it very clear it's the label for a question. And it's consistent with other fields which don't have the `question*` prefix, and the UI which also call these just `label`.

src/main/resources/db/migration/V1_25_add_question_label_to_questions.sql
```diff
@@ -1,2 +1,2 @@
-ALTER TABLE question ADD COLUMN question_label VARCHAR(60);
-ALTER TABLE historical_question ADD COLUMN question_label VARCHAR(60);
+ALTER TABLE question ADD COLUMN question_label TEXT;
```
xoen, line 1: The column name also has to match, e.g. being called `label` (not `question_label`). I suppose you can also simplify the name of the migration to `*_add_label_to_questions.sql`.

Could we also add the label to responses? Have we checked with Michael what's the intent of this change? Presumably it's to take a sort of snapshot of what was asked/answered (so including the response)

Mjwillis (commented):
Some re-working here, but a good start

src/main/resources/db/migration/V1_25__add_question_label_to_questions.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE question ADD COLUMN label TEXT;
+ALTER TABLE historical_question ADD COLUMN label TEXT;
+
+update question set label = question;
+update historical_question set label = question;
+
+ALTER TABLE question ALTER COLUMN label SET NOT NULL;
+ALTER TABLE historical_question ALTER COLUMN label SET NOT NULL;
```
xoen, line 8: Instead of altering the column created in this PR, I'd just change the `ADD COLUMN` statements above to add `NOT NULL` columns instead.

(Same result but tidier migration)

Mjwillis (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/Response.kt
```diff
@@ -12,6 +12,8 @@ data class Response(
   val code: String?,
   @Schema(description = "The response text as seen by downstream data consumers")
   val response: String,
+  @Schema(description = "The response text as seen by the user at the point of entry", nullable = true)
```
xoen, line 15: Another `nullable = true` that should go.

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/AddOrUpdateQuestionWithResponses.kt
```diff
@@ -26,6 +26,13 @@ data class AddOrUpdateQuestionWithResponses(
   )
   @field:Size(min = 1)
   val question: String,
+  @Schema(
+    description = "The question text as seen by the user at the point of entry",
+    requiredMode = Schema.RequiredMode.NOT_REQUIRED,
```
xoen, line 31: What does `requiredMode` do? If the field is mandatory should this be `NOT_REQUIRED`? 🤔

PR #357, opened 2025-07-08
IR-1098, IR-1100 & IR-1115: Increase test coverage of new label & code fields

ushkarev:
… and ensure that NOMIS sync updates the question and response `label`s in case the text changes!


src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/NomisSyncResourceTest.kt
```diff
@@ -748,11 +786,6 @@ class NomisSyncResourceTest : SqsIntegrationTestBase() {
               """
               {
                 "reportReference": "112414666",
-                "type": "SELF_HARM_1",
```
ushkarev, line 751: the test above is already very thorough with field checks so making this one specific purely to descriptions

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/service/NomisSyncServiceTest.kt
```diff
@@ -285,14 +290,17 @@ class NomisSyncServiceTest {
       initialMigration = initialMigration,
     )
 
-    whenever(reportRepository.save(any())).thenReturn(sampleReport)
```
ushkarev, line 288: having the repository return `sampleReport` meant that calling `assertSampleReportConvertedToDto` further down did not in fact test the result of the upsert!

instead, the mocked save method now returns the real newly-created report.

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
```diff
@@ -3225,29 +3245,6 @@ class ReportResourceTest : SqsIntegrationTestBase() {
           )
         }
 
-        @Test
-        fun `can update existing question with responses that contain codes in response`() {
```
ushkarev, line 3229: this is now redundant because responses must always have codes

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/jpa/Question.kt
```diff
@@ -103,6 +103,7 @@ class Question(
         newResponse?.let { newVersion ->
           this.responses.find { it == newVersion }?.apply {
             response = newVersion.response
+            label = newVersion.label
```
ushkarev, line 106: label must be overwritten as well in case of response text change

PR #370, opened 2025-07-31
Adding duplicated report ID to report table, update report call, and …

jonesj2193:
…both basic and 'with details' report objects

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/UpdateReportRequest.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/UpdateReportRequest.kt
@@ -7,6 +7,8 @@ import jakarta.validation.constraints.Size
 import uk.gov.justice.digital.hmpps.incidentreporting.constants.InformationSource
 import uk.gov.justice.digital.hmpps.incidentreporting.jpa.Report
 import java.time.LocalDateTime
+import java.util.Optional
+import java.util.UUID
 
 @Schema(
   description = "Payload to update key properties of an incident report",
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/dto/request/UpdateReportRequest.kt
```diff
@@ -95,6 +105,7 @@ data class UpdateReportRequest(
     description?.let { report.description = it }
     staffInvolvementDone?.let { report.staffInvolvementDone = it }
     prisonerInvolvementDone?.let { report.prisonerInvolvementDone = it }
+    duplicatedReportId?.let { report.duplicatedReportId = it }
```
ushkarev, line 108: i think u need a `.getOrNull()` after the it

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
```diff
@@ -1048,6 +1063,7 @@ class ReportResourceTest : SqsIntegrationTestBase() {
           location = "LEI",
           title = "Updated report 11124143",
           description = "Updated incident report of type find of illicit items",
+          duplicatedReportId = Optional.of(UUID.fromString(duplicatedExistingReport.id.toString())),
```
ushkarev, line 1066: out of curiosity, isn’t `duplicatedExistingReport.id` already a UUID? could this be:
```kotlin
Optional.of(duplicatedExistingReport.id)
```

(similar below)

ushkarev (dismissed):

src/test/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResourceTest.kt
ushkarev: we know that once saved the report must have an id so you could
`Optional.of(duplicatedExistingReport.id!!)`

ushkarev (approved):

PR #380, opened 2025-08-14
Renovate tweaks

xoen:
- ~wait 3 days before using a new version of a dependency~
- bump `swagger-parser` separately
- bump DPR library separately

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -15,6 +15,14 @@
     },
     {
       "matchManagers": ["gradle"],
+      "matchPackageNames": ["hmpps-digital-prison-reporting-lib"],
+      "matchUpdateTypes": ["minor", "patch"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -3,12 +3,30 @@
...
+      "matchManagers": ["gradle"],
+      "matchPackageNames": ["swagger-parser"],
+      "matchUpdateTypes": ["minor", "patch"],
+      "groupName": "swagger-parser",
+      "groupSlug": "swagger-parser",
+      "minimumReleaseAge": "3 days"
+    },
+    {
+      "matchManagers": ["gradle"],
+      "matchPackageNames": ["hmpps-digital-prison-reporting-lib"],
```
ushkarev, line 18: do u wanna separate this in order to protect against non-major-version breaking changes? not against this at all, but that was much more of an issue in the ui app

PR #396, opened 2025-08-28
IR-1171 Align Helm values file configurations

Mjwillis:
- Standardized memory requests and limits across all environments.
- Updated `JAVA_OPTS` in configurations where previously missing.
- Adjusted production memory limits to align with development and preproduction settings.

### DIFF
--- a/helm_deploy/hmpps-incident-reporting-api/values.yaml
+++ b/helm_deploy/hmpps-incident-reporting-api/values.yaml
@@ -10,8 +10,10 @@ generic-service:
   replicaCount: 4
 
   resources:
+    requests:
+      memory: 4G
     limits:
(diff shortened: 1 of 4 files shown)


helm_deploy/hmpps-incident-reporting-api/values.yaml
```diff
@@ -41,7 +43,7 @@ generic-service:
           return 401;
         }
   env:
-    JAVA_OPTS: "-Xmx3g"
+    JAVA_OPTS: "-Xmx4g"
```
xoen, line 46: you could probably go a bit higher given the k8s limit is now 6GB but let's see how this goes.

xoen (approved):

PR #554, opened 2026-04-20
IR-1582: Hyphenate Self-harm

xoen:
This requires a migration as the descriptions for
types and type families are also in the DB.

NOTE: This may have cause newly generated Self-harm reports' titles to be different than before. But there is already a precedent where a typo in a type name was changed (Food refusal) so it should be ok (See: https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/415#discussion_r2417016792)


jonesj2193 (approved):

Mjwillis (approved):
LGTM

PR #563, opened 2026-05-12
IR-1613 Remove NOMIS validation; add "Unlawful Detention" incident

Mjwillis:
### Summary

- Removed NOMIS validation logic and related constants.
- Added new incident type, "Unlawful Detention," along with associated family and roles configurations.

### Code changes

The following sections summarize the major updates:
- **Validation:** Completely removed NOMIS validation logic and any constants tied to it.
- **Incident Types:** Introduced "Unlawful Detention" as a new incident type, with its corresponding family and roles added.

---

### Checklist

- [ ] Code changes have been updated in relevant documentation.
- [ ] Unit tests added/updated to cover these changes.
- [ ] All tests pass successfully in CI/CD pipelines.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/NoDetailsGiven.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/NoDetailsGiven.kt
@@ -1,4 +1,0 @@
-package uk.gov.justice.digital.hmpps.incidentreporting.constants
-
-/** Replacement text when some information was not provided by NOMIS */
-const val NO_DETAILS_GIVEN = "NO DETAILS GIVEN"
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/PrisonerOutcome.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/PrisonerOutcome.kt
@@ -1,7 +1,5 @@
 package uk.gov.justice.digital.hmpps.incidentreporting.constants
 
-import jakarta.validation.ValidationException
-
 /**
  * The outcome of a prisoner’s involvement in an incident.
  *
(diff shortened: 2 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/NoDetailsGiven.kt
```diff
@@ -1,4 +0,0 @@
-package uk.gov.justice.digital.hmpps.incidentreporting.constants
-
-/** Replacement text when some information was not provided by NOMIS */
-const val NO_DETAILS_GIVEN = "NO DETAILS GIVEN"
```
xoen, line 4: I don't remember how this was doing, but presumably it's not used anywhere?

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/PrisonerOutcome.kt
```diff
@@ -32,10 +30,4 @@ enum class PrisonerOutcome(
   SEEN_OUTSIDE_HOSP("Seen by outside hospital", "OUTH"),
   TRANSFER("Transfer", "TRN"),
   TRIAL("Trial", "TRL"),
-  ;
-
-  companion object {
-    fun fromNomisCode(code: String): PrisonerOutcome = entries.find { it.nomisCode == code }
```
xoen, line 38: Interesting, we weren't using these `fromNomisCode()` methods?

src/main/resources/db/migration/V1_43__unlawful_detention_constants.sql
```diff
@@ -0,0 +1,9 @@
+-- add new type
+insert into constant_type_family(sequence, code, description)
+values ((select max(sequence)+1 from constant_type_family),  'UNLAWFUL_DETENTION', 'Unlawful detention');
```
xoen, line 3: tbh we could probably hardcode the sequences as they're fairly predictable and kinda reads better (and helps when adding future migrations as you don't need to look into the DB)

xoen (approved):
Looks good to me.

I think I'd prefer predictable sequences in the migrations as we've done in the past, maybe change it as it would be consistent with past migrations.

src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/constants/PrisonerOutcome.kt
Mjwillis: not any more

src/main/resources/db/migration/V1_43__unlawful_detention_constants.sql
Mjwillis: True

PR #568, opened 2026-05-19
IR-1638 Update report logic and add filtering by prisoner role

Mjwillis:
### What does this pull request do?

- Updates the specification and count logic to handle distinct rows in joins.
- Increases the page size limit to 1000 needed for other services


src/main/kotlin/uk/gov/justice/digital/hmpps/incidentreporting/resource/ReportResource.kt
```diff
@@ -253,8 +253,8 @@ class ReportResource(
     @PageableDefault(page = 0, size = 20, sort = ["incidentDateAndTime"], direction = Sort.Direction.DESC)
     pageable: Pageable,
   ): SimplePage<ReportBasic> {
-    if (pageable.pageSize > 50) {
-      throw ValidationException("Page size must be 50 or less")
+    if (pageable.pageSize > 1000) {
```
Mjwillis, line 256: 50 was too restrictive for other services

xoen (approved):

